PR #327, opened 2024-05-21
remove dependency on google cloud storage for gpu drivers

gainsley:
Changes the handling of GPU driver files to remove dependency on GCS (Google Cloud Storage), which on-prem deployments may not want to depend on.

Instead of creating a GPU Driver where the Controller downloads the driver file and uploads to our own GCS, GPU Driver files are now treated the same as VM images, and expected to be hosted somewhere already.
- The GPU Driver is a URL to either a publicly available file, or to our internal artifact registry, where the user has already uploaded the file.
- The license config is treated the same way - it is no longer the contents of the license file, but a URL to the license file, either publicly available, or already uploaded to the artifact registry.

This leverages the mechanisms and storage we already have with the artifact registry, which itself could be backed by GCS or S3 or others in the future.


levshvarts (commented):
Looks good - much simpler to follow what's happening; however license config is usually not a publicly available file - hence we were storing it as text.

gainsley:
Yup, so the flow would be to upload the license config file into the VM registry somewhere and point it from the GPUDriver config.

levshvarts:
Ah, gotcha, so same process as any artifact.

levshvarts (approved):

PR #328, opened 2024-05-28
remove hard-coded image refs

gainsley:
The code had hard-coded references to images:
- hard-coded version for vm base image (could be overridden per cloudlet, but we want default defined by the deployment)
- hard-coded reference to envoy-with-curl and nginx-with-curl images.

The full references to the images are now taken as command line arguments to CCRM, which then puts them into PlatformConfig, which is then used to generate CRM command line arguments.


pkg/ccrm/ccrm.go
```diff
@@ -59,6 +59,8 @@ type Flags struct {
 	DnsZone                       string
 	CloudletRegistryPath          string
 	CloudletVMImagePath           string
+	EnvoyWithCurlImage            string
```
levshvarts, line 62: I would suggest just renaming them to envoyImage and NginxImage to be more generic

pkg/platform/common/infracommon/common-platform.go
```diff
@@ -147,7 +147,8 @@ func GetPlatformConfig(cloudlet *edgeproto.Cloudlet, pfConfig *edgeproto.Platfor
 		Region:              pfConfig.Region,
 		TestMode:            pfConfig.TestMode,
 		CloudletVMImagePath: pfConfig.CloudletVmImagePath,
-		VMImageVersion:      cloudlet.VmImageVersion,
+		EnvoyWithCurlImage:  pfConfig.EnvoyWithCurlImage,
+		NginxWithCurlImage:  pfConfig.NginxWithCurlImage,
```
levshvarts, line 151: Same comment as before - just `EnvoyImage` and `NginxImage`

levshvarts (approved):
Looks good. Just a clarification on base image version. Right now if we set `VmImageVersion` for the cloudlet to `5.1.0` for instance it will try to boot pf/sharedLb with edgecloud-5.1.0 image. Is my understanding correct that now we need to set `VmImageVersion` to `edgecloud-5.1.0`?

gainsley:
> Looks good. Just a clarification on base image version. Right now if we set `VmImageVersion` for the cloudlet to `5.1.0` for instance it will try to boot pf/sharedLb with edgecloud-5.1.0 image. Is my understanding correct that now we need to set `VmImageVersion` to `edgecloud-5.1.0`?

Thanks for reviewing, Lev! `vmImageVersion` isn't being used anymore. `cloudletVmImagePath` is the full path to the file, including any version name as part of the directory path or the filename. We used to do this song-and-dance where the path root, the filename, and the version were all specified separately and concatenated together, but that was too confusing. There's now just the full path, and the md5sum.

pkg/ccrm/ccrm.go
gainsley: Hm, but wouldn't that invite confusion? If someone specifies the latest upstream envoy, things are going to fail because it doesn't have curl built into it. At least with this name it's clear that this is not just a vanilla envoy image.

levshvarts: From the code, it looks like we use curl for unix domain socket case only, which is an e2e/unit test case. We can keep the "withCurl" note, since we are using it for now, and if we change it at some point not to need curl at all, we can change the var names.

Issue #331, opened 2024-05-30
Heat stack failed to update on cluster scale up


levshvarts:
When a customer tried to update cluster to bump up number of worker nodes, the update failed with the following error:

```
Encountered failures: update failed: error udpating heat stack: berlin-main-ts-cluster-lab5gbrawo-heat.yaml -- ERROR: The specified reference "mex-k8s-master-berlin-main-ts-cluster-lab5gbrawo-shared-volume" (in mex-k8s-master-berlin-main-ts-cluster-lab5gbrawo.Properties.block_device_mapping[0].volume_id) is incorrect. , exit status 1
```

It looks like crm didn't merge the old heat stack and the new one upon generation.

levshvarts:
Fixed.

PR #334, opened 2024-05-30
Fixed shared volume merging on heat stack update.

levshvarts:
### Issues Fixed

[* Heat stack failed to update on cluster scale up](https://github.com/edgexr/edge-cloud-platform/issues/331)

### Description

Shared volumes were treated as being part of the VM spec, which they shouldn't be. Moved the block for shared volumes, so when we re-use the VM template we still add the shared volume block.

### DIFF
--- a/pkg/platform/common/vmlayer/vmparams.go
+++ b/pkg/platform/common/vmlayer/vmparams.go
@@ -587,6 +587,9 @@ type VMOrchestrationParams struct {
 	AuthPublicKey           string
 	DeploymentManifest      string
 	Command                 string
+	// TODO - Volumes should be just a reference here and
+	// and the volume definition should be in VMGroupOrchestrationParams
+	// similar how ports and handled.
 	Volumes                 []VolumeOrchestrationParams
 	Ports                   []PortResourceReference      // depending on the orchestrator, IPs may be assigned to ports or
 	FixedIPs                []FixedIPOrchestrationParams // to VMs directly
--- a/pkg/platform/openstack/openstack-heat.go
+++ b/pkg/platform/openstack/openstack-heat.go
@@ -602,7 +602,7 @@ func (o *OpenstackPlatform) populateParams(ctx context.Context, VMGroupOrchestra
 		}
 	}
 
-	// Get chef keys for existing VMs
+	// Get existing VMs definition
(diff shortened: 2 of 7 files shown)


pkg/platform/openstack/openstack-heat.go
```diff
@@ -178,6 +175,9 @@ resources:
             availability_zone: {{.AvailabilityZone}}
             {{- end}}
     {{- end}}
+	{{- if .ExistingData }}
```
gainsley, line 178: So this rebuilds the volume definition from the new state, but another way to do this would be to include the existing volume definition in the ExistingData object. The choice depends on if there is there a way that 1) the volume definition could change (like the size) or 2) the volumes could be removed/added. If yes, then does that trigger a rebuild of the VM? I do see that it looks like changing the volumes may affect the cloud-init config, which would definitely trigger a VM rebuild.

Looking at clusterinst.proto, it looks like external_volume_size and shared_volume_size are not updatable, so technically volumes should not change. I think then it might be safer to leave the template as it was and just ensure that ExistingData includes any volume definitions. What do you think?

levshvarts: I tested and since the attachment VM never gets updated, the volume stays the same and doesn't get blown away(stays with the same ID). I think this is ok, however let me explore keeping it as part of the Existing data to see, which approach is cleaner

levshvarts: After looking at how to handle the volumes separately from the VMs it is not straightforward. We should address this in another [issue](https://github.com/edgexr/edge-cloud-platform/issues/335)

pkg/platform/common/vmlayer/vmparams.go
```diff
@@ -587,6 +587,9 @@ type VMOrchestrationParams struct {
 	AuthPublicKey           string
 	DeploymentManifest      string
 	Command                 string
+	// TODO - Volumes should be just a reference here and
+	// and the volume definition should be in VMGroupOrchestrationParams
+	// similar how ports and handled.
```
levshvarts, line 592: Tracked in [this issue](https://github.com/edgexr/edge-cloud-platform/issues/335)

PR #336, opened 2024-05-31
Updated NoSchedule taint name for master

levshvarts:
### Description
`node-role.kubernetes.io/master:NoSchedule` taint has been replaced with `node-role.kubernetes.io/control-plane:NoSchedule` from k8s 1.20.

### DIFF
--- a/pkg/k8smgmt/cluster.go
+++ b/pkg/k8smgmt/cluster.go
@@ -54,10 +54,10 @@ func SetMasterNoscheduleTaint(ctx context.Context, client ssh.Client, masterName
 	var cmd string
 	if action == NoScheduleMasterTaintAdd {
 		log.SpanLog(ctx, log.DebugLevelInfra, "adding taint to master", "masterName", masterName)
-		cmd = fmt.Sprintf("kubectl taint nodes %s node-role.kubernetes.io/master=:NoSchedule --kubeconfig=%s", masterName, kubeconfig)
+		cmd = fmt.Sprintf("kubectl taint nodes %s node-role.kubernetes.io/control-plane=:NoSchedule --kubeconfig=%s", masterName, kubeconfig)
 		out, err := client.Output(cmd)
 		if err != nil {
-			if strings.Contains(out, "already has node-role.kubernetes.io/master") {
+			if strings.Contains(out, "already has node-role.kubernetes.io/control-plane") {
 				log.SpanLog(ctx, log.DebugLevelInfra, "master taint already present")
 			} else {
 				log.SpanLog(ctx, log.DebugLevelInfra, "error adding master taint", "out", out, "err", err)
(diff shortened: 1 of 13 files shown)


gainsley (commented):
Should we be tainting with both in case it's an old cluster? Or this only happens for new cluster creates and we're definitely using v20 there? Maybe we should be checking the kubernetes version and then using the correct taint?

levshvarts:
> Should we be tainting with both in case it's an old cluster? Or this only happens for new cluster creates and we're definitely using v20 there? Maybe we should be checking the kubernetes version and then using the correct taint?

You are right, we should be checking the version - old clusters were created with `node-role.kubernetes.io/master:NoSchedule`, so if we are updating a cluster and need to move workload onto a master tainting should be use the old annotations.
My concern is how to handle these issues  going forward, since we'll be hitting changes like this periodically.
@gainsley - what do you think?

gainsley:
I think we need to track the cluster k8s version anyway, to allow for upgrade (triggered by users), and to be able to notify users that the cluster version is old/deprecated and needs to be upgraded. If we are tracking the k8s version, then we can determine the correct taint to use.

It looks like we must use the old taint before 1.20, and we must use the new taint from 1.24. Sources:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/kubeadm/2067-rename-master-label-taint
https://kubernetes.io/blog/2022/04/07/upcoming-changes-in-kubernetes-1-24/#api-removals-deprecations-and-other-changes-for-kubernetes-1-24

levshvarts:
Updated code to check for older k8s cluster version we had. I made it explicit, since that was the only version we had as far as I recall.

pkg/k8smgmt/cluster.go
```diff
@@ -52,12 +55,22 @@ func SetMasterNoscheduleTaint(ctx context.Context, client ssh.Client, masterName
...
 
 	var cmd string
+	taintAnnotation := NoScheduleMasterTaintAnnotaionNew
+	cmd = fmt.Sprintf("kubectl version --short --kubeconfig=%s | grep Server", kubeconfig)
+	out, err := client.Output(cmd)
+	if err != nil {
+		return fmt.Errorf("Unable to get k8s version, %v", err)
+	}
+	// For old clusters NoSchedule taint is different
+	if out == "Server Version: v1.18.20" {
```
gainsley, line 65: Not sure if you tested this against a v1.18 cluster but I might recommend using strings.Contains() to allow for `\n` and other extra stuff that might show up in the output.

gainsley (approved):

pkg/k8smgmt/cluster.go
levshvarts: Good idea. I didn't test it out against 1.18 just because we don't have one that'd be easy to test.

PR #338, opened 2024-06-01
workarounds intermittent docker login and openstack failures

gainsley:
- docker login from new LB to pull envoy (proxy) container failed to contact harbor but worked later. Allow for retry. Previously the envoy container was hosted publicly and did not need docker login.
- openstack cli commands fail intermittently with `Failed to discover available identity versions when contacting https://bdgecloud.t.c:5000/v3. Attempting to parse version from URL.`. Allow for retry to avoid these intermittent failures causing the whole VM/cluster deploy process to fail.


levshvarts (approved):
Looks good, but does it only fail on the first run? What's the percentage of two failed commands? Especially openstack error - that seems like a high chance of two consecutive errors.

gainsley:
I don't have any stats on failures. That might be a good thing to track, though. I can add some stats accumulation to the code.

PR #342, opened 2024-07-09
Re-build shared rootLb state 

levshvarts:
### Description
We hit an issue, where shared rootLb heat stack was stuck in Update state. To fix this we need to delete the stack, which will trigger the deletion of the shared rootLb VM, which is a problem, since we don't have a way to rebuild the state on the newly created rootLb VM.

Added code to re-build shared rootLb state, if the previous one was deleted. When the crm comes up it does one of the two things - updates rootLb VM, or creates it. If shared rootLb was created, we check if there are any clusters on this cloudlet, that use the shared rootLb, and if there are several things are done(`CheckRebuildRootLb`):
1. Update rootLb vm with the cluster network
2. update rootLb vm with kubeconfig for the cluster
3. Walk the appInsts to rebuild manifest states for all the created appInsts on the shared cloudlet.

In addition updated `InitHAConditional` to return whether rootLb was updated, or created.

### DIFF
--- a/pkg/platform/common/vmlayer/appinst.go
+++ b/pkg/platform/common/vmlayer/appinst.go
@@ -370,6 +370,7 @@ func (v *VMPlatform) setupDnsForAppInst(ctx context.Context, clusterInst *edgepr
 			if err != nil {
 				return err
 			}
+		} else {
 			patchIPV4 = sips.IPV4ExternalAddr()
 			patchIPV6 = sips.IPV6ExternalAddr()
 		}
--- a/pkg/platform/common/vmlayer/appinst.go
+++ b/pkg/platform/common/vmlayer/appinst.go
@@ -316,6 +316,84 @@ func seedDockerSecrets(ctx context.Context, client ssh.Client, clusterInst *edge
 	return nil
 }
 
+func (v *VMPlatform) setupDnsForAppInst(ctx context.Context, clusterInst *edgeproto.ClusterInst, app *edgeproto.App, appInst *edgeproto.AppInst, client ssh.Client, deployment string) error {
+	var proxyOps []proxy.Op
+	var getDnsAction func(svc v1.Service) (*infracommon.DnsSvcAction, error)
+
+	rootLBName := v.VMProperties.GetRootLBNameForCluster(ctx, clusterInst)
+	rootLBIPs, err := v.GetExternalIPFromServerName(ctx, rootLBName, WithCachedIP(true))
+	if err != nil {
+		return err
+	}
+	names, err := k8smgmt.GetKubeNames(clusterInst, app, appInst)
+	if err != nil {
+		return err
+	}
+	sips, err := v.GetIPFromServerName(ctx, v.VMProperties.GetCloudletMexNetwork(), v.GetClusterSubnetName(ctx, clusterInst), GetClusterMasterName(ctx, clusterInst))
+	if err != nil {
+		return err
+	}
+	ops := infracommon.ProxyDnsSecOpts{AddProxy: true, AddDnsAndPatchKubeSvc: true, AddSecurityRules: true}
+	wlParams := infracommon.WhiteListParams{
+		SecGrpName:  infracommon.GetServerSecurityGroupName(rootLBName),
+		ServerName:  rootLBName,
+		Label:       infracommon.GetAppWhitelistRulesLabel(app),
+		AllowedCIDR: infracommon.GetAllowedClientCIDR(),
+		Ports:       appInst.MappedPorts,
+		DestIP:      infracommon.DestIPUnspecified,
+	}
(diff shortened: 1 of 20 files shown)


pkg/platform/common/vmlayer/vmprovider.go
```diff
@@ -560,6 +562,120 @@ func (v *VMPlatform) initRootLB(ctx context.Context, platformConfig *platform.Pl
...
+	err = pc.CreateDir(ctx, client, configDir, pc.NoOverwrite, pc.NoSudo)
+	if err != nil {
+		log.SpanLog(ctx, log.DebugLevelInfra, "unable to create a directory", "AppInst", appInst.Key, "err", err)
+		return
+	}
+	file := configDir + "/" + configName
+	log.SpanLog(ctx, log.DebugLevelInfra, "writing config file", "file", file, "kubeManifest", mf)
+	err = pc.WriteFile(client, file, mf, "K8s Deployment", pc.NoSudo)
+	if err != nil {
+		log.SpanLog(ctx, log.DebugLevelInfra, "unable to write manifest", "AppInst", appInst.Key, "err", err)
```
gainsley, line 626: This same set of code is duplicated in k8smgmt.createOrUpdateAppInst() (although it's actually missing the network policy part). I would suggest you break out the existing code in k8smgmt.createOrUpdateAppInst() into a sub function like EnsureDeploymentManifest(), so that it can be called from both there and here, to avoid duplicating code. The new function should be idempotent. Also that way you don't need to export GetConfigDirName()/GetConfigFileName(), as it remains an implementation detail of the k8smgmt package.

pkg/platform/common/vmlayer/vmprovider.go
```diff
@@ -488,13 +489,14 @@ func (v *VMPlatform) InitHAConditional(ctx context.Context, platformConfig *plat
 			}
 		}()
 	} else {
+		rootLbAction = platform.ActionCreate
 		err := v.initRootLB(ctx, platformConfig, ActionCreate, updateCallback)
 		if err != nil {
-			return err
+			return platform.ActionNone, err
```
gainsley, line 495: Is there any reason we shouldn't just call CheckRebuildRootLb() right here after initRootLB() is called for the first time? I don't really see any need for it to be run from the CRM. Only if we manually deleted the rootLB will it incur any extra real delay, which will almost never be the case. Not all platforms may necessarily have rootLBs, either.

gainsley (commented):
Thanks for fixing this!

pkg/platform/common/vmlayer/vmprovider.go
levshvarts: We could yes(initially I thought caches wouldn't be ready there, but they got initialized way before the call to initHAConditional) - we'll just need to pass caches to InitHAConditional(), which I didn't want to do, but it makes sense - this way there is no need to make checkRebuildRootLb() part of the platform interface. Let me change that.

gainsley: Cool, thanks. And you don't need to pass caches to InitHAConditional, they're already stored on VMPlatform.Caches as part of InitCommon().

pkg/platform/common/vmlayer/vmprovider.go
levshvarts: With respect to network policy - I didn't add it, since we don't create multi-tenant clusters, but probably a good idea to add it, since we support them.

levshvarts: actually, this is something that should only be done if creating a new cluster, but in this case, we are re-building state, so network policy would already be added to the cluster.

gainsley (approved):
Looks good!

pkg/platform/common/vmlayer/appinst.go
```diff
@@ -316,6 +316,84 @@ func seedDockerSecrets(ctx context.Context, client ssh.Client, clusterInst *edge
...
+	switch deployment {
+	case cloudcommon.DeploymentTypeHelm:
+		fallthrough
+	case cloudcommon.DeploymentTypeKubernetes:
+		patchIPV4 := ""
+		patchIPV6 := ""
+		if v.VMProperties.GetUsesMetalLb() {
+			patchIPV4 = sips.IPV4ExternalAddr()
+			patchIPV6 = sips.IPV6ExternalAddr()
+		}
```
gainsley, line 361: Shouldn't this be if not uses metal lb?

gainsley (approved):
Nice, great to reduce all that duplicate code.

pkg/platform/common/vmlayer/appinst.go
levshvarts: Thanks! Good catch - I was looking at the logic to make sure I don't mess it up...and still did :)

Issue #340, opened 2024-06-18
Cannot create cloudlet in a restricted mode


levshvarts:
When you try to create a cloudlet with a restricted access, you get an error: "Cloudlet key {} not found", which seems to come from node access api.

levshvarts:
Fixed.

PR #346, opened 2024-07-23
Chef server references cleanup.

levshvarts:
### Issues Fixed
https://github.com/edgexr/edge-cloud-platform/issues/340

### Description

Restricted cloudlet was failing to generate manifest due to leftover chef key references. Cleaned up chef references to address the failures as well as get rid of the dead code.


gainsley (approved):
Nice, thanks for fixing this!

pkg/platform/common/vmlayer/cloudlet.go
```diff
@@ -89,9 +89,28 @@ func (v *VMPlatform) GetPlatformNodes(cloudlet *edgeproto.Cloudlet) []NodeInfo {
...
+		log.SpanLog(ctx, log.DebugLevelInfra, "using image from MEX_OS_IMAGE property", "imgFromProps", imgFromProps)
+		return imgFromProps, nil
+	}
+
+	vmBaseImage := v.VMProperties.CommonPf.PlatformConfig.CloudletVMImagePath
+	if vmBaseImage == "" {
+		return "", fmt.Errorf("Get cloudlet image failed, cloudletVMImagePath not set")
+	}
+	imageNameWithoutExt := util.RemoveExtension(filepath.Base(vmBaseImage))
+	return imageNameWithoutExt, nil
```
gainsley, line 107: this code is duplicated on lines 115-124, I suggest you remove those lines and replace them with this new function.

gainsley (approved):
looks good!

pkg/platform/common/vmlayer/cloudlet.go
levshvarts: So I was thinking about this, but the logic becomes more complicated in `GetCloudletImageToUse`. But adding an extra return value made the code ok-ish.

pkg/platform/common/vmlayer/cloudlet.go
```diff
@@ -89,21 +89,38 @@ func (v *VMPlatform) GetPlatformNodes(cloudlet *edgeproto.Cloudlet) []NodeInfo {
...
+func (v *VMPlatform) GetCloudletImageToUse(ctx context.Context, updateCallback edgeproto.CacheUpdateCallback) (string, error) {
+	imageNameWithoutExt, vmBaseImage, err := v.GetCloudletImageName(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	// means we use image from props
+	if vmBaseImage == "" {
+		return imageNameWithoutExt, nil
+	}
```
gainsley, line 123: You're right I was confused for a while why we need two strings. But now I think I get it. I suggest you rename `vmBaseImage` to `imagePath`. The distinction (which wasn't clear right away) is that one is just the name, and the other is the full path. Then I suggest changing this check to:
```
// if no image path, nothing to download, so just return the image name
if imagePath == "" {
    return imageNameWithoutExt, nil
}
```

PR #347, opened 2024-07-30
Added startup FQDN/URI fields to appinst, clusterinst, and cloudlet.

levshvarts:
### Description
In order to accomodate DNS migration for platform we needed to add immutable startup fields, which will be used to find the correct VM names, since those are derived from the FQDN names.

Added fields to clusterinst.proto, appinst.proto and cloudlet.proto to manage startup names.

Added upgrade function for etcd objects in order to set the new startup fields correctly(initially these will get set to runtime values of fqdn)


gainsley (approved):
Looks good! Instead of "startup", can we use something that denotes that it should never change. Maybe "static" or "immutable" or "const"?

levshvarts:
> Looks good! Instead of "startup", can we use something that denotes that it should never change. Maybe "static" or "immutable" or "const"?

I think "startup" denotes that it's something that doesn't changes - since that's what it was at startup. Other names don't convey where this value gets set at. Do you feel another prefix is better?

gainsley:
I don't know, "startup" just sounds like something that happened at startup, but everything happens at startup and doesn't speak to whether it can be modified or not. To me "startup" is kinda like "initial". I would prefer something that implies it doesn't change, same a constant.

levshvarts:
changed `startup` to `static`

api/edgeproto/appinst.proto
```diff
@@ -134,6 +134,8 @@ message AppInst {
   distributed_match_engine.Loc cloudlet_loc = 3 [(gogoproto.nullable) = false];
   // Base FQDN (not really URI) for the App. See Service FQDN for endpoint access.
   string uri = 4 [(protogen.backend) = true];
+  // Static startup FQDN that was used when appinst was created
```
gainsley, line 137: `that was used when appinst was created` suggest: "used" -> "set"

api/edgeproto/cloudlet.proto
```diff
@@ -578,6 +578,8 @@ message Cloudlet {
   string dns_label = 54 [(protogen.backend) = true];
   // Root LB FQDN which is globally unique
   string root_lb_fqdn = 55 [(protogen.backend) = true];
+  // Root LB FQDN which was set at startup and cannot be changed
```
gainsley, line 581: `set at startup` sounds like it could be referring to startup of process/service, I suggest `set when object was created`, like you have for AppInst

PR #348, opened 2024-07-30
grpc connection cache

gainsley:
This adds a GRPC connection cache to manage multiple GRPC clients. This is actually somewhat copied from similar code in MC, which manages GRPC clients per region.

I'm adding it to edge-cloud-platform so we can manage GRPC clients per CCRM nodeType from the Controller.

### DIFF
--- a/pkg/cloudcommon/grpc_conncache.go
+++ b/pkg/cloudcommon/grpc_conncache.go
@@ -98,15 +98,15 @@ func (s *GRPCConnCache) SetConn(key string, conn *grpc.ClientConn) {
 func (s *GRPCConnCache) Run(ctx context.Context) {
 	s.Lock()
 	defer s.Unlock()
-	for region, conn := range s.cache {
-		used := s.used[region]
+	for key, conn := range s.cache {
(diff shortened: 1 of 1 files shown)


pkg/cloudcommon/grpc_conncache.go
```diff
@@ -0,0 +1,120 @@
...
+func (s *GRPCConnCache) SetConn(key string, conn *grpc.ClientConn) {
+	s.Lock()
+	defer s.Unlock()
+	s.cache[key] = conn
+}
+
+func (s *GRPCConnCache) Run(ctx context.Context) {
+	s.Lock()
+	defer s.Unlock()
+	for region, conn := range s.cache {
```
levshvarts, line 101: Maybe rename this to something more generic, like `key`.

levshvarts (approved):
Looks good.

pkg/cloudcommon/grpc_conncache.go
gainsley: good point, will do

PR #349, opened 2024-07-30
spanlog grpc logging, no-log-start-finish

gainsley:
This adds additional logging capabilities to our GRPC interceptors, allowing for better debuggability on both client and server. I also added a new span option to avoid logging to disk the "start" and "finish" logs of the span. This is for quick go threads that need their own span but don't want to be noisy printing the start/stop messages.


levshvarts (approved):
Looks good.

PR #350, opened 2024-07-31
cloudlet ssh key provider

gainsley:
This adds a cloudlet ssh key provider which provides on-demand signed keys. This is intended to replace the code in `pkg/platform/common/infracommon/common-ssh.go` which forces each platform instance to maintain a pair of ssh keys, which require a persistent thread to refresh the signed keys every 24h. And this ssh key is always set up on platform init.

With the intent to move to CRM functionality into the CCRM, the platform instance needs to be ephermal and created on demand. The current ssh key handling means that, in the CCRM:
- we may instantiate a platform to run an API call which doesn't need any ssh keys
- we may instantiate a platform that doesn't actually use ssh keys (public cloud)
- we have no way of keeping a persistent thread set up by the platform around

So instead, this library allows for replacing the forced, fixed ssh keys with an on-demand one. It only initializes keys when needed and doesn't need a refresh thread - it can detect if the current key is expired and refresh it automatically.

The other thing we will do here is share this ssh key signer between all platforms, so new platforms in the CCRM will not pay any overhead for setting up ssh keys.

### DIFF
--- a/pkg/platform/common/cloudletssh/sshkey.go
+++ b/pkg/platform/common/cloudletssh/sshkey.go
@@ -37,6 +37,8 @@ type SSHKey struct {
 	mux               sync.Mutex
 	expiresAt         time.Time
 	refreshInProgress bool
+	refreshErr        error
+	refreshWait       sync.WaitGroup
 }
 
 // KeySigner is used to sign the user's public key
@@ -46,7 +48,8 @@ type KeySigner interface {
 }
 
 var SignSSHKeyTimeout = 30 * time.Second
-var RefreshBufferDuration = 2 * time.Hour
+var RefreshInlineBufferDuration = time.Minute
+var RefreshLazyBufferDuration = 2 * time.Hour
 
 func NewSSHKey(signer KeySigner) *SSHKey {
 	return &SSHKey{
@@ -64,7 +67,7 @@ func (s *SSHKey) getSignedKey(ctx context.Context) (string, time.Time, error) {
 	if err != nil {
 		return "", time.Time{}, fmt.Errorf("failed to sign cloudlet ssh key, %s", err)
 	}
-	// set expiration time
+	// determine expiration time
 	pk, err := xssh.ParsePublicKey([]byte(signedKey))
 	if err != nil {
 		return "", time.Time{}, fmt.Errorf("failed to parse signed cloudlet key, %s", err)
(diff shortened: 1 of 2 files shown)


pkg/platform/common/cloudletssh/sshkey.go
```diff
@@ -0,0 +1,145 @@
...
+	log.SpanLog(ctx, log.DebugLevelInfra, "Sign cloudlet public key from Vault")
+
+	ctx, cancel := context.WithTimeout(ctx, SignSSHKeyTimeout)
+	defer cancel()
+
+	signedKey, err := s.signer.SignSSHKey(ctx, s.publicKey)
+	if err != nil {
+		return "", time.Time{}, fmt.Errorf("failed to sign cloudlet ssh key, %s", err)
+	}
+	// set expiration time
```
levshvarts, line 67: shouldn't this be `// get expiration time` ?

pkg/platform/common/cloudletssh/sshkey.go
```diff
@@ -0,0 +1,145 @@
...
+		return "", time.Time{}, fmt.Errorf("failed to parse signed cloudlet key, %s", err)
+	}
+	cert := pk.(*xssh.Certificate)
+	expiresAt := time.Unix(int64(cert.ValidBefore), 0)
+	return signedKey, expiresAt, nil
+}
+
+func (s *SSHKey) ensureSignedKeyLocked(ctx context.Context) error {
+	now := time.Now()
+	if s.signedPublicKey == "" || now.After(s.expiresAt) {
```
levshvarts, line 79: don't we need to also check `refreshInProgress`? Otherwise we can potentially fire off `getSignedKey` here and in a go routine

pkg/platform/common/cloudletssh/sshkey.go
```diff
@@ -0,0 +1,145 @@
...
+		if err != nil {
+			return err
+		}
+		s.signedPublicKey = signedKey
+		s.expiresAt = expiresAt
+	} else if now.After(s.expiresAt.Add(-RefreshBufferDuration)) && !s.refreshInProgress {
+		// current key is still valid but expiring soon, spawn
+		// thread to refresh it
+		s.refreshInProgress = true
+		go func() {
```
levshvarts, line 92: I feel like a channel approach would work better here, since really the mutex is only needed once(and in a go routine, but this is where the channel might make it better).

Idea is to have a `refresh` channel member in `SSHKey` and have `getSignedKey` write to that channel for synchronization. 

What do you think?

pkg/platform/common/cloudletssh/sshkey_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/edgexr/edge-cloud-platform/pkg/log"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/crypto/ssh"
+)
+
+type keySigner struct {
+	signer      ssh.MultiAlgorithmSigner
+	signedCount int
+	validDur    time.Duration
+}
```
levshvarts, line 36: I guess our primary signer would be vault-based. Are you thinking to implement that in a later PR?

levshvarts (commented):
Nice to have a more on-demand way of signing the certs. I had some questions in-line.

pkg/platform/common/cloudletssh/sshkey.go
gainsley: i'll clarify it

pkg/platform/common/cloudletssh/sshkey.go
gainsley: I guess there could have been a go thread spawned by a call right before the cert expired, then another call comes in here after the cert expired and the go thread is still running. Let me see if I can handle that case.

pkg/platform/common/cloudletssh/sshkey.go
gainsley: Hm we still need the mutex to be able to read the key in GetKeyPairs (it's not atomic). Also for the go thread case, nobody would actually be reading from the channel, since the current call isn't waiting for the new cert (current cert is still valid and can be used).

pkg/platform/common/cloudletssh/sshkey_test.go
gainsley: platform.AccessApi already implements the KeySigner interface from sshkey.go, so we already have a Vault-based implementation. The test code here is just emulating what Vault does internally to sign a cert.

gainsley:
Hey Lev, I've addressed your review comments. I couldn't think of a way to use a channel that would help - note that there may be multiple threads waiting for a single refresh go thread to finish. To address the issue with the inline refresh ignoring a potential refresh thread already running, now the inlne refresh is also done using a go thread, we just need to wait for that thread to finish.

levshvarts (approved):
Nice - thanks for making the changes. Looks good.

PR #351, opened 2024-07-31
add ansible shepherdlb role

gainsley:
This adds a new ansible role for shepherdLB, which is a configuration where shepherd and prometheus run on the LB, and there is no on-edge-site CRM (so the LB needs to have the accesskeys for Shepherd to auth directly against the controller).

Some tasks are moved from a specific role to the common templates directory.

This isn't in use as of yet.


levshvarts (approved):
Looks good.

PR #352, opened 2024-07-31
move etcd code to regiondata package

gainsley:
This change moves etcd-related code out of pkg/controller and into its own common package, "regiondata". This will allow the etcd code to be leveraged by the CCRM service as well as the Controller. There are no functional changes here (except to allow multiple watchers in the dummy etcd server). Everything else is just moving code around and dealing with renames.


levshvarts (approved):
Looks good.

gainsley:
Thanks Lev.

PR #353, opened 2024-07-31
syncdata for synchronizing reservations

gainsley:
A common pattern in the platform code is to the need to compute a new subnet/IP to use. The way it does this is to call an infra API to see what subnets/IPs are currently in use, decide on a new free one, and then use it by writing it as part of another infra API call. This may happen in parallel in different threads. This is a common write-after-read pattern. Any easy fix to avoid two threads trying to use the same value is wrap this work with a mutex.

In cases like Openstack, the "writing" part is part of a Heat stack that is doing everything - creating VMs, networks, security groups, etc. It takes a long time (minutes), and any failure would cause the whole thing to be rolled back. So obviously we don't want to wrap the whole thing with a mutex. The fix then is to only wrap the calculation of the free value under a mutex and store it in memory. Release the lock, let the long-running write to infra happen, and after it has finished, grab the mutex again and release the reservation. All threads must check both the reservations in memory along with the state of the infra to determine a new free value.

The above works fine in the context of the CRM, because it is the single and only process that is interacting with the infra. However, we want to be able to allow the CCRM, a set of horizontally scaled services, to run the same type of code (for HA among other reasons). So we cannot use a simple mutex.

The solution is to follow the same procedure, but rather than using a mutex and an in-memory table, we use a transactional database (like Etcd) to store the reserved values. There are two deviations from the mutex based approach:
1) we track an object ownerID with each reserved value. If the CCRM crashes in the middle of work, the Controller will still have the owner object (i.e. AppInst/ClusterInst) in it's database. We can either retry the work, or if the object is deleted by the user, we can clean up the reservation in Etcd via the ownerID. This avoids the issue of a stale reservation.
2) we do not drop the reservation after the work is done (we could actually, as the source of truth is really what is stored in the infra). But we would still need to clean up any entries when the owner object is deleted to avoid stale entries, so it's redundant to do both.

This code introduces a "syncdata" package which defines a SyncFactory interface. A SyncFactory instance allows the platform code to create new SyncReservation objects which provides for doing the patterns above. The platform doesn't actually know the storage backing of the Factory and its SyncReservation objects. In the context of the CRM, it will be mutex based, in the context of the CCRM, it will Etcd based (note that CRM doesn't have access to Etcd).

The "syncdata" package provides the generic interfaces, and the mutex-based implementation. The "regiondata" package provides the etcd-backed implementation.


levshvarts (approved):
Looks good - the interface looks intuitive. So the plan is to use etcd as kvstore backend as I understand. I recall we had an issue with transient data exploding the memory footprint of etcd. I suspect we might hit the same issue with a sufficient number of transactions. What do you think?

gainsley:
Yup, we'll use etcd as the kvstore, since unfortunately redis isn't stable storage, and if it gets shutdown we could potentially leak stale reservations. I don't think it will be an undue burden on etcd's storage (it was disk, not memory if I recall), as the rate of these transactions is based on changes to ClusterInsts/AppInsts, which should be low compared to things like Alerts (which we still store in etcd, btw). But it's a good thing to keep an eye on.

Issue #337, opened 2024-05-31
Consolidate external network configuration during cloudlet creation


levshvarts:
We have a common way of specifying external network, which can be set with [`external_network_name`](https://github.com/edgexr/edge-cloud-platform/blob/8b4b546611ccaa1d0b7feb76be1174fd24d0ca4b/api/edgeproto/cloudlet.proto#L256). However crm only looks at `VMProviderProps` env vars for the network to be used in the creation of rootLB VMs(including shared VM). The result of this is that if only `external_network_name` is set, only the platform VM will be created with the correct network, but the sharedLB VM will be created with the default `external-network-shared` network, which is the default. This can lead to the confusing results as well as make the crm incapable to communicate with the rootLBs, if the networks are prevented to directly communicated. 

Was this the original intent, or is this a bug? In any case, we should make sure both the platform VM and rootLB instances are created on the same network.

gainsley:
I remember we did have a case where one Operator wanted the CRM on a separate protected network (control plane) from the rest of the VMs (data plane) and the CRM was behind some kind of NAT/firewall. I don't remember the specifics. But if that were the case here, the variable names here should have reflected that, which they don't. So not sure either what's going on here.

PR #355, opened 2024-08-02
allow crm code to run from ccrm

gainsley:
Previously, our architecture required a CRM service running on edge-site to convert platform-independent APIs to platform-specific API calls to deploy VMs/Clusters/AppInsts/etc. To be able to support platforms where it is not feasible to run the CRM service on edge-site (typically because there is another layer of software managing the infrastructure), we want to be able to run the platform-specific code from the CCRM service, which runs off-edge-site alongside the Controller.

This PR refactors the common CRM code (from `pkg/crmutil/controller-data.go`), moving the parts that are specific to a single CRM on-edge into `pkg/crm`. The remaining code is modified to support being included in either the CRM, which is a single-instance process using notify for communication with the Controller, or the CCRM which is a horizontally scaled set of processes that use direct access to etcd and GRPC to communicate with the Controller. Code in `pkg/crmutil` is meant to be shared between both CRM and CCRM.

In the common `pkg/crmutil/controller-data.go` code, there was a lot of functionality that was specific to a single-instance process running over notify. The following changes were made:
- The functions in controller-data.go now use a "InfoSender" object to send data back to the controller. In the CRM case, this writes to caches which trigger sending over notify as before. In the CCRM case, this writes directly back to the GRPC API.
- The "vmResourceActionEnd" trigger that used to run after each ClusterInst/AppInst is now moved to the CRM or CCRM.
- All of the code in controller-data.go related to applying TrustPolicyExceptions is rewritten in `pkg/controller/trustpolicyexception_changes.go`.
- The VMPool-related code is moved only to the CRM. It is written in such a way that it cannot support being done from the CCRM, and I did not want to rewrite it since no one is using it. So VMPool is only support for CRM on-edge.

The platform code also requires changes to be able to run from CCRM. This was only partially completed.
- Platforms that use reservations of subnets/IPs can do so via the SyncFactory
- The cert refresh thread which used to be part of the VMProvider code is now moved to the CRM `pkg/crm/crm.go`. For the CCRM, the Controller has a periodic thread that runs to trigger CCRMs to refresh certs in `pkg/controller/ccrm_periodic.go`.
- The cloudlet ssh key refresh thread is replaced by the CloudletSSH package, which provides on-demand signed ssh certs.
TODO:
- Some platforms use mutexes to prevent multiple threads from modifying the infra at the same time. These need to be handled such that things safely work if multiple processes are modifying the infra at the same time.
- The VMProvider code has a lock to protect against simultaneous writing to a VM (network config files, IP tables rules). This can probably be replaced by using file locks to get a fencing token on read, and then the same file lock to abort a write if the fencing token is stale (same concept as etcd transactions). Note the file lock is not actually held between the read and the write.

Other changes:
- The fieldMap is now its own object instead of a map[string]string. I have changed the way a field flag for a field that is a struct is interpreted, i.e. CloudletInfoFieldStatus, which refers to the Status (edgeproto.StatusInfo) field of edgeproto.CloudletInfo. Previously, setting the Status field doesn't imply anything on its own, but it was required if a subfield of StatusInfo was set, i.e. CloudletInfoFieldStatusTaskName. The CopyInFields() function used to require both to be set to copy in the TaskName, but I have changed it such that if the parent CloudletInfoFieldStatus flag is set, it copies in the entire struct. I have also updated the way the hand-written code uses flags, as sometimes it was checking if the flag or its parent was set, or it was really trying to check if the flag or any of its children were set. The FieldMap code is now auto-generated and ends up in `api/edgeproto/alert.pb.go`.

I still need to run tests against a real infra like Openstack, and maybe take care of the AppInstRuntime TODO.


pkg/ccrm/ccrm_cloudlet.go
```diff
@@ -40,65 +43,40 @@ var (
...
-	var errState edgeproto.TrackedState
-	var successState edgeproto.TrackedState
-	var workFunc func(context.Context, *edgeproto.Cloudlet, platform.Platform, edgeproto.CacheUpdateCallback) error
+	responseSender := edgeproto.NewCloudletInfoSendUpdater(ctx, stream, in.Key)
 
-	if in.OnboardingState == edgeproto.TrackedState_CREATE_REQUESTED {
-		ackState = edgeproto.TrackedState_CREATING
-		errState = edgeproto.TrackedState_CREATE_ERROR
-		successState = edgeproto.TrackedState_READY
+	var workFunc func(context.Context, *edgeproto.Cloudlet, platform.Platform, *edgeproto.CloudletInfoSendUpdater) error
```
levshvarts, line 55: might be better to define with worker function type, so the variable declaration is cleaner looking, but this is a nit.

pkg/ccrm/ccrm_cloudlet.go
```diff
@@ -143,36 +112,64 @@ func (s *CCRMHandler) createCloudlet(ctx context.Context, in *edgeproto.Cloudlet
...
-func (s *CCRMHandler) deleteCloudlet(ctx context.Context, in *edgeproto.Cloudlet, cloudletPlatform platform.Platform, cb edgeproto.CacheUpdateCallback) error {
+	log.SpanLog(ctx, log.DebugLevelApi, "ccrm setup new cloudlet", "cloudlet", in)
+	// replicate here what CRM would have done to set up initial cloudlet state
+	// Note that "cloudletPlatform" is initialized in CreateCloudlet, similar to
+	// but not the same as platform.InitCommon(), so we need to get a different
+	// copy of the platform object that has been initialized with InitCommon.
+	pf, err := s.getCRMCloudletPlatform(ctx, &in.Key)
+	if err != nil {
+		if errors.Is(err, ErrPlatformNotFound) {
+			return nil
```
levshvarts, line 128: Should we log that the platform is of unknown type?

gainsley: yup, I'll add a log

gainsley:
Hi Lev, I found a bunch of issues after more testing. In particular, I didn't realize the proxyCerts object was called by the infra-specific code, so I needed to split it into a stateless and cloudlet-specific part, and a persistent and cloudlet-independent part (cache).

api/edgeproto/cloudlet.proto
```diff
@@ -516,8 +522,8 @@ message Cloudlet {
   PlatformConfig config = 21 [(gogoproto.nullable) = false, (protogen.backend) = true];
   // Optional resource to restagtbl key map key values = [gpu, nas, nic]
   map<string, edgeproto.ResTagTableKey> res_tag_map = 22 [(protogen.backend) = true, (protogen.refers_to) = "ResTagTable"];
-  // Secrets required to access cloudlet, will be saved in encrypted storage
-  map<string, string> access_vars = 23;
+  // Secrets required to access cloudlet, will be saved in encrypted storage, and cleared on the cloudlet object
+  map<string, string> access_vars = 23 [(protogen.backend) = true, (protogen.hidetag) = "nocmp"];
```
levshvarts, line 526: Doesn't adding `(protogen.backend) = true` make this not configurable?

gainsley: No, it just removes it from comparisons in unit/e2e tests. "noconfig" makes it not configurable by the user.

levshvarts: Yeah, I see the corresponding code in mex.go. So comment in `protogen.proto` is just incorrect:
```
  // Backend defines a field that is filled in by the back-end,
  // not by the user.
  bool backend = 51013;
```

gainsley: Yeah, it's a little misleading. I think at one point it was meant to be as you said, only configurable by the back-end, as opposed to how it's treated now, which is *may* be configured by the back-end.

pkg/ccrm/ccrm_appinst.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2024 EdgeXR, Inc
```
levshvarts, line 1: Since this is the implementation of `AppInstPlatformAPI`, does it make sense to call the files `ccrm_appinst_api.go` to be consistent with all the implementations of the APIs defined in `.proto` files? Same question applies to cloudlet and clusterinst

gainsley: We could, although I don't think we need to. Everything in here is for the APIs anyway.

pkg/ccrm/ccrm_handler.go
```diff
@@ -55,29 +76,112 @@ type NodeAttributesData struct {
...
+	s.crmHandler.GPUDriverCache.InitSync(sync)
+	s.crmHandler.AlertPolicyCache.InitSync(sync)
+	s.crmHandler.NetworkCache.InitSync(sync)
+	s.caches.PlatformFeaturesCache.InitSync(sync)
+	s.caches.CloudletNodeCache.InitSync(sync)
+	s.caches.CloudletInfoCache.InitSync(sync)
+	s.caches.ClusterInstInfoCache.InitSync(sync)
+	s.caches.AppInstInfoCache.InitSync(sync)
+	nodeMgr.CloudletLookup.GetCloudletCache(node.NoRegion).InitSync(sync)
+	nodeMgr.CloudletPoolLookup.GetCloudletPoolCache(node.NoRegion).InitSync(sync)
```
levshvarts, line 131: Nit - I know that region is not used here, since we are using region cache, but just to make it clear, it's probably a good idea to pass the region(`nodeMgr.Region`).

pkg/ccrm/ccrm_handler.go
```diff
@@ -55,29 +76,112 @@ type NodeAttributesData struct {
...
+		edgeproto.RegisterAppInstPlatformAPIServer(grpcServer, s)
+	}
+}
+
+func (s *CCRMHandler) Start(ctx context.Context, ctrlConn *grpc.ClientConn) {
+	s.ctrlConn = ctrlConn
+	for _, builder := range s.platformBuilders {
+		plat := builder()
+		features := plat.GetFeatures()
+		features.NodeType = s.nodeMgr.MyNode.Key.Type
```
levshvarts, line 153: `fatures.NodeType = s.nodeType`
otherwise, what's the point of an extra member field in CCRMHandler :)

pkg/ccrm/ccrm_cloudletapi.go
```diff
@@ -144,19 +142,32 @@ func (s *CCRMHandler) NameSanitize(ctx context.Context, in *edgeproto.NameSaniti
...
-	if key == nil {
-		return nil, nil, fmt.Errorf("CloudletKey not specified")
+func (s *CCRMHandler) RefreshCerts(ctx context.Context, in *edgeproto.Cloudlet) (*edgeproto.Result, error) {
+	pf, err := s.getCRMCloudletPlatform(ctx, &in.Key)
+	if err != nil {
+		return nil, err
+	}
+	features := pf.GetFeatures()
+
+	proxyCerts := certs.NewProxyCerts(ctx, &in.Key, pf, s.nodeMgr, nil, features, s.flags.CommercialCerts, s.flags.EnvoyWithCurlImage, s.proxyCertsCache)
```
levshvarts, line 152: Why not keep the proxyCerts as part of the CCRMHandler and just  start the refresh thread here instead of delegating this to ccrm_periodic?

pkg/ccrm/ccrm_exec.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func (s *CCRMHandler) ProcessExecRequest(ctx context.Context, in *edgeproto.CloudletExecReq) (*edgeproto.ExecRequest, error) {
+	_, platform, err := s.getCloudletPlatform(ctx, in.CloudletKey)
+	if err != nil {
+		return nil, err
+	}
+	resp := make(chan *edgeproto.ExecRequest, 1)
+
+	go func() {
```
levshvarts, line 32: What's a point of spawning a go routine, if we just sit on the channel recv waiting for that function to finish?

pkg/ccrm/ccrm_cloudletapi.go
gainsley: For load balancing. ccrm_periodic distributes the refresh for each cloudlet to different CCRMs. If we just did it here, this CCRM would have to run the refresh for every cloudlet.

pkg/ccrm/ccrm_handler.go
gainsley: Good point :)

pkg/ccrm/ccrm_exec.go
gainsley: Good point, I probably just copied it from the crm code. I'll unwrap it.

pkg/cloudcommon/grpc.go
```diff
@@ -127,3 +131,65 @@ var GRPCClientKeepaliveParams = keepalive.ClientParameters{
...
+// the error to a string.
+func GRPCErrorUnwrap(err error) error {
+	st, ok := status.FromError(err)
+	if ok && st != nil {
+		return errors.New(st.Message())
+	}
+	return err
+}
+
+type GRPCStreamRecv[Object any] interface {
```
levshvarts, line 146: I thought we'd be able to not use generics for a bit longer :)

pkg/controller/appinst_api.go
```diff
@@ -1236,12 +1239,39 @@ func (s *AppInstApi) createAppInstInternal(cctx *CallContext, in *edgeproto.AppI
 	reqCtx, reqCancel := context.WithTimeout(ctx, s.all.settingsApi.Get().CreateAppInstTimeout.TimeDuration())
 	defer reqCancel()
 
-	err = edgeproto.WaitForAppInstInfo(reqCtx, &in.Key, s.store,
-		edgeproto.TrackedState_READY,
-		CreateAppInstTransitions, edgeproto.TrackedState_CREATE_ERROR,
-		"Created AppInst successfully", cb.Send,
-		edgeproto.WithCrmMsgCh(sendObj.crmMsgCh),
-	)
+	crmAction := func() error {
```
levshvarts, line 1242: I am not sure why we need to define `crmAction`, since we just call it right away - this can just be inlined here.

pkg/controller/appinst_api.go
```diff
@@ -1866,11 +1936,38 @@ func (s *AppInstApi) deleteAppInstInternal(cctx *CallContext, in *edgeproto.AppI
 		reqCtx, reqCancel := context.WithTimeout(ctx, s.all.settingsApi.Get().DeleteAppInstTimeout.TimeDuration())
 		defer reqCancel()
 
-		err = edgeproto.WaitForAppInstInfo(reqCtx, &in.Key, s.store, edgeproto.TrackedState_NOT_PRESENT,
-			DeleteAppInstTransitions, edgeproto.TrackedState_DELETE_ERROR,
-			"Deleted AppInst successfully", cb.Send,
-			edgeproto.WithCrmMsgCh(sendObj.crmMsgCh),
-		)
+		crmAction := func() error {
```
levshvarts, line 1939: Same comment - why not inline the function, since it just gets called right away?

pkg/ccrm/ccrm_cloudletapi.go
levshvarts: ah, ok. Otherwise all ccrm instances will run this thread. Makes sense.

pkg/cloudcommon/grpc.go
gainsley: They're actually pretty useful :) It's either this or I need to define some new protogen options and auto-generate this code from the protobufs, which means I basically write this same code with templates instead of generics. I actually prefer the generics as VSCode understands them and can detect syntax/compile errors.

The only drawback of generics is you can't use generic functions on an object, which is somewhat limiting. But the object can be generic.

pkg/controller/appinst_api.go
gainsley: It's to simplify the error handling. You see there's a lot of error paths there, and we need to be able to ignore any of them if IGNORE_CRM_ERRORS is set.

pkg/controller/cloudlet_api.go
```diff
@@ -983,23 +1049,24 @@ func (s *CloudletApi) UpdateCloudlet(in *edgeproto.Cloudlet, inCb edgeproto.Clou
 	if err != nil {
 		return err
 	}
-	if !ignoreCRMState(cctx) {
```
levshvarts, line 986: Is this check not needed anymore?

pkg/controller/clusterinst_api.go
```diff
@@ -460,41 +455,35 @@ func (s *ClusterInstApi) GetRootLBFlavorInfo(ctx context.Context, stm concurrenc
 }
 
 // getAllCloudletResources
-// Returns (1) All the VM resources on the cloudlet (2) Diff of VM resources reported by CRM and seen by controller
-func (s *ClusterInstApi) getAllCloudletResources(ctx context.Context, stm concurrency.STM, cloudlet *edgeproto.Cloudlet, cloudletInfo *edgeproto.CloudletInfo, cloudletRefs *edgeproto.CloudletRefs) ([]edgeproto.VMResource, []edgeproto.VMResource, bool, error) {
+// Returns (1) All the VM resources on the cloudlet
```
levshvarts, line 458: nit - don't need `(1)` anymore, since it only returns one thing anyways

pkg/controller/appinst_api.go
levshvarts: Ok, I see.

pkg/controller/trustpolicyexception_api.go
```diff
@@ -26,28 +26,30 @@ import (
...
+	edgeproto.InitTPEInstanceStateCache(&trustPolicyExceptionApi.instCache)
 	sync.RegisterCache(&trustPolicyExceptionApi.cache)
+	sync.RegisterCache(&trustPolicyExceptionApi.instCache)
 	return &trustPolicyExceptionApi
 }
 
 func (s *TrustPolicyExceptionApi) CreateTrustPolicyException(ctx context.Context, in *edgeproto.TrustPolicyException) (*edgeproto.Result, error) {
 
 	log.SpanLog(ctx, log.DebugLevelApi, "CreateTrustPolicyException", "policy", in)
-	if len(in.OutboundSecurityRules) == 0 {
```
levshvarts, line 48: why is this check not required anymore?

pkg/controller/trustpolicyexception_changes_test.go
```diff
@@ -0,0 +1,273 @@
...
+
+	"github.com/edgexr/edge-cloud-platform/api/edgeproto"
+	"github.com/edgexr/edge-cloud-platform/pkg/cloudcommon/node"
+	"github.com/edgexr/edge-cloud-platform/pkg/log"
+	"github.com/edgexr/edge-cloud-platform/pkg/regiondata"
+	"github.com/edgexr/edge-cloud-platform/test/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+// Tests for TPEInstancState
```
levshvarts, line 30: nit - TPEInstancState - TPEInstanceState

pkg/platform/common/vmlayer/lb.go
```diff
@@ -49,6 +49,7 @@ var udevRulesFile = "/etc/udev/rules.d/70-persistent-net.rules"
 var sharedRootLBPortLock sync.Mutex
 
 var RootLBPorts = []dme.AppPort{}
+var NoAccessKey string
```
levshvarts, line 52: Not used anywhere.

pkg/platform/common/vmlayer/ssh.go
```diff
@@ -215,133 +215,3 @@ func (v *VMPlatform) GetAllCloudletVMs(ctx context.Context, caches *platform.Cac
 	log.SpanLog(ctx, log.DebugLevelInfra, "GetAllCloudletVMs done", "cloudletVMs", fmt.Sprintf("%v", cloudletVMs))
 	return cloudletVMs, nil
 }
-
-func GetVaultCAScript(publicSSHKey string) string {
```
levshvarts, line 219: Nice to get rid of the legacy upgrade code.

pkg/platform/common/vmlayer/vmprovider.go
```diff
@@ -408,38 +400,30 @@ func (v *VMPlatform) InitCommon(ctx context.Context, platformConfig *platform.Pl
...
-	v.FlavorList, err = v.VMProvider.GetFlavorList(ctx)
-	if err != nil {
-		log.SpanLog(ctx, log.DebugLevelInfra, "GetFlavorList failed", "err", err)
-		return err
-	}
 	var cloudlet edgeproto.Cloudlet
 	if !v.Caches.CloudletCache.Get(v.VMProperties.CommonPf.PlatformConfig.CloudletKey, &cloudlet) {
 		return fmt.Errorf("unable to find cloudlet key in cache")
 	}
+	// TODO: resolve why there are two ways to store/use the external network name
```
levshvarts, line 420: https://github.com/edgexr/edge-cloud-platform/issues/337

pkg/platform/common/vmlayer/vmprovider.go
```diff
@@ -408,38 +400,30 @@ func (v *VMPlatform) InitCommon(ctx context.Context, platformConfig *platform.Pl
...
 	v.VMProperties.PlatformExternalNetwork = cloudlet.InfraConfig.ExternalNetworkName
+	if cloudlet.InfraConfig.ExternalNetworkName != "" {
```
levshvarts, line 422: This change essentially fixes https://github.com/edgexr/edge-cloud-platform/issues/337 since now infraconfig takes precedence.

pkg/platform/openstack/openstack-heat.go
```diff
@@ -404,37 +405,58 @@ func IsUserDataSame(ctx context.Context, userdata1, userdata2 string) bool {
...
 
-	if len(VMGroupOrchestrationParams.Subnets) > 0 {
+	ownerID := VMGroupOrchestrationParams.OwnerID
+	// reserve function may run several times if there are conflicts,
+	// so track which CIDRs need to be updated so they can be re-assigned
+	// on subsequent runs.
+	subnetNeedsResource := make(map[int]struct{})
+	for i, s := range VMGroupOrchestrationParams.Subnets {
+		if s.CIDR == vmlayer.NextAvailableResource {
+			subnetNeedsResource[i] = struct{}{}
```
levshvarts, line 424: Do we need just one here? We can probably break after finding the first available.

pkg/platform/openstack/openstack-heat.go
```diff
@@ -583,70 +611,80 @@ func (o *OpenstackPlatform) populateParams(ctx context.Context, VMGroupOrchestra
...
 		if f.FloatingIpId == vmlayer.NextAvailableResource {
-			var fipid string
-			var err error
-			if action == heatTest {
-				fipid = "test-fip-id"
-			} else {
-				fipid, err = o.getFreeFloatingIpid(ctx, VMGroupOrchestrationParams.Netspec.FloatingIPExternalNet)
-				if err != nil {
-					return nil, err
+			ipNeedsResource[i] = struct{}{}
```
levshvarts, line 625: Same here - should we break if we just need a single one?

pkg/platform/vcd/vcd.go
```diff
@@ -107,12 +115,6 @@ func (v *VcdPlatform) InitProvider(ctx context.Context, caches *platform.Caches,
 		}
 		log.SpanLog(ctx, log.DebugLevelInfra, "InitProvider", "mexInternalNetRange", mexInternalNetRange)
 
-		log.SpanLog(ctx, log.DebugLevelInfra, "InitProvider update isonet metadata", "stage", stage)
-		err = v.UpdateLegacyIsoNetMetaData(ctx)
```
levshvarts, line 111: Is this really not used anymore? Since we don't have vcd setup to test, how can we be sure?

pkg/process/cloudlet_services.go
```diff
@@ -269,6 +281,51 @@ func StopCRMService(ctx context.Context, cloudlet *edgeproto.Cloudlet, haRole HA
 	return nil
 }
 
+var maxPrimaryCrmStartupWait = 10 * time.Second
+
+func StartCRMServicesHA(ctx context.Context, cloudlet *edgeproto.Cloudlet, pfConfig *edgeproto.PlatformConfig, redisCfg *rediscache.RedisConfig, crmOps ...CrmServiceOp) error {
```
levshvarts, line 286: We never test this on a real platform. Do you think we actually still need HA CRMs? If we are moving to CCRM providing HA, this feels like throwaway code, or something that will just accumulate technical debt.

pkg/process/process_elasticsearch.go
```diff
@@ -38,9 +39,12 @@ func (p *ElasticSearch) StartLocal(logfile string, opts ...StartOp) error {
 
 func (p *ElasticSearch) StartElasticSearch(logfile string, opts ...StartOp) error {
 	// simple single node cluster
+	if p.Port == "" {
```
levshvarts, line 42: We don't run anything else on 9200. Is this something specific to your setup, or do we not clean something up properly?

tools/protoc-gen-gomex/mex.go
```diff
@@ -2504,13 +2815,20 @@ func (m *mex) generateMessage(file *generator.FileDescriptor, desc *generator.De
...
+		}
 		m.cacheTemplate.Execute(m.gen.Buffer, args)
 		m.importUtil = true
 		m.importLog = true
 		if args.WaitForState != "" {
 			m.importErrors = true
 			m.importStrings = true
 		}
+		if args.ParentObjName != "" {
+			m.importSync = true
```
levshvarts, line 2830: I am thinking if overloading the meaning of `parent_obj_name` option might have implications in the future...  I am thinking if we will have some state data that's not explicitly coming from cloudlet, which is managed by CCRM/CRM. What do you think?

levshvarts (approved):
Phew, that was a gargantuan set of changes. Thank you for taking this on.
I guess this is the next step - to test `RequiresCrmOnEdge: false` with openstack. I see that this is set to true for now in for openstack as well as for vcd. Did you test with fake platform just to validate the functionality?

pkg/platform/vcd/vcd.go
gainsley: There's this comment in vcd-network.go: `// on (re)start attempt ensure legacy (prior to common shared lb) ISO nets are in metadata`, so I'm pretty sure this isn't relevant anymore.

pkg/platform/common/vmlayer/vmprovider.go
gainsley: It's still confusing to me why we have two ways (env var and cloudlet.infraconfig) of specifying the same thing. But if I helped at least address some issue here, that's great :)

pkg/process/cloudlet_services.go
gainsley: There is still the restricted cloudlet case, where we would need it if HA is needed. I agree it's not well tested and not likely to be used anyway, but (like some other features :) I didn't want to rip it out just because it's inconvenient. I think if we discuss and agree there's no place for it (i.e. we don't support HA for restricted cloudlets, which could be reasonable), then we can rip it out. Let's discuss.

pkg/platform/openstack/openstack-heat.go
gainsley: This is just tracking which subnets need to be assigned a reservable resource. The issue here is that the resource allocation function overwrites the "NextAvailableResource" value when it assigns a resource. But then that function may fail to commit, in which case that function gets run again. But the second time around, s.CIDR's value may not be "NextAvailableResource", it may be what was assigned on the previous run that did not commit. So here we're just tracking which subnets need CIDR's assigned so that the reservation function can run multiple times ok.

pkg/process/process_elasticsearch.go
gainsley: So I was (and still am) seeing intermittent failures in the TestEvents unit test. It always passes when I run it standalone, but it fails in `make unit-test`. I suspected that multiple unit tests are trying to write events. Normally these just fail silently, but if they overlap with the TestEvents test, which is actually checking the state of events, it might add extra junk which causes the test to fail. That was my hypothesis anyway, so I though if I changed the port for the TestEvents test, the other tests wouldn't be able to write to ES. Unfortunately i think I still see that test failing sometimes.

tools/protoc-gen-gomex/mex.go
gainsley: I don't think I'm overloading it here. It's meant to identify the relationship between the "info" and its parent, i.e. CloudletInfo and its parent Cloudlet. And CloudletInfo is always what's used to send back information to the Controller in response to a Cloudlet change, which is the same thing the existing WaitFor code is doing, and the new code I added is doing (just over GRPC instead of notify). I agree "parent_obj_name" is not really an accurate name in terms of sending responses (although that's the primary purpose of the "info" structs). If anything I would rename it something like `response_obj_for`.

gainsley:
Thanks Lev also for taking on that code review! So I have done a wide range of tests, besides the usual unit and e2e tests, I also set it up (you can see this in the director changes) to run `make test-start-dns`, which let me create a cloudlet on Openstack from a CRM started via e2e tests. I also tested a k3s deployment with the operator changes on Openstack (both those openstack tests used the acceptance tests to test). But yeah, I only tested openstack with CrmOnEdge=true. All the CrmOnEdge=false testing is via the fake platform in unit/e2e tests.

I think that when we implement the OSM platform with CrmOnEdge=false we'll have a chance to find any other bugs lurking there.

pkg/platform/common/vmlayer/vmprovider.go
levshvarts: I think this changes the behavior, but probably we should retire one way altogether.

pkg/platform/openstack/openstack-heat.go
levshvarts: Ah ok, now that makes sense.

pkg/process/process_elasticsearch.go
levshvarts: hm, ok. Yeah, since multiple unit-tests will run in parallel its hard to figure out how to properly address this, but if this works, that's fine.

pkg/ccrm/ccrm_exec.go
gainsley: Actually Lev, we do actually need the go routine. The ProcessExecReq() function needs to send back the response before it completes, and then block to keep the pipes open to read/write from the shell process (for an interactive shell). Once the shell exits, then the ProcessExecReq() exists. Alternatiavely, ProcessExecReq may fail before running the interactive shell, in which case it doesn't block, and we send back the failure. I'll add some comments.

pkg/controller/cloudlet_api.go
gainsley: Yeah, I couldn't think of why this check was here, it didn't make any sense. I tried looking through the history, I think a bunch of the CRM upgrade code used to be bundled under this check, but it got moved out and only the access vars update was left behind. Does it make sense to you?

pkg/controller/trustpolicyexception_api.go
gainsley: I moved that check into the in.Validate() function, so it's still there. The problem was the Update function wasn't checking if the user had removed all the rules, so now Validate() checks for that on both Create and Update.

pkg/platform/common/vmlayer/lb.go
gainsley: It's still used in calls to GetVMSpecForRootLB in common/vmlayer/appinst.go and clusterinst.go. I do see it's not actually used in the func, so now I changed the check in GetVMSpecForRootLB to compare accessKey to `NoAccessKey` instead of `""`.

pkg/controller/cloudlet_api.go
levshvarts: makes sense. but if feels like the check is there not to update accessvars in the vault if IgnoreCrm is set. I don't think this matters either way, as CRM is not in the picture here, so yeah this checks looks like it's not doing anything.

PR #356, opened 2024-08-12
avoid conflicting kubeconfig names

gainsley:
I noticed a bug in the way we generate kubeconfig names. To avoid aliasing, we now use the ObjId field (which is a ULID). This guarantees universal uniqueness, although the file name now lacks any human-recognizable name.

This change only applies to ClusterInsts created after this change goes in.


levshvarts (commented):
Good catch - I am surprised this has never come up! I am concerned about it not being human-readable though, we do use those names if we need to manually debug the cluster. What about just adding org as part of the name as well?

gainsley:
> I am concerned about it not being human-readable though, we do use those names if we need to manually debug the cluster. What about just adding org as part of the name as well?

Yeah, I'm also concerned about that. Unfortunately it kind of negates the point of this change if we append potentially mutable data to the immutable string. For example if we append the org name, but in the future add a feature to allow the org to be renamed, then we need to add another backwards-compatible check here. I think we only very rarely directly use these kubeconfig files manually, so I don't think it's a big issue here. What do you think?

gainsley:
Going to rework the changes as part of the cloudlet key change.

PR #357, opened 2024-08-15
remove cloudletkey from instance keys

gainsley:
This removes the CloudletKey from the AppInstKey and ClusterInstKey, allowing for a separation of concerns from the AppInst/ClusterInst keys (owned by the developer) and the Cloudlet keys (owned by the operator). The main reason for this is for the upcoming Zones feature, which intends to hide Cloudlets from developers by putting Cloudlets in a new Zone object.  This also moves us closer to the CAMARA APIs which use a single string ID as the unique identifier for objects. So now the AppInstKey and ClusterInstKeys only have a single unique string "name" (although scoped to the organization for multi-tenancy).

Due to the changes of the keys:
- ClusterInstKey is removed, as it has become equivalent to the existing ClusterKey.
- AppInstRefKey is removed, as it is now equal to the new AppInstKey
- I've removed older upgrade functions, as they are not needed anymore and would require work to rewrite using the new objects.

The bulk of the changes here are mechanical, moving the CloudletKey from the AppInstKey to the AppInst body, and from the ClusterInstKey to the ClusterInst body. No functional change, as the same information is preserved.

Some functional changes were required in places where we were passing just the instance key around, but now the key doesn't have the cloudletKey information, and these places didn't have easy access to the full instance body that now contains the cloudletKey info.

The remaining functional changes are to deal with the change in scope of the AppInst/ClusterInst key. Previously the key only needed to be unique within the scope of a single cloudlet (CloudletKey). Now, the instance key must be unique within the region. This requires renaming instances that have conflicts during upgrade. It also then requires being able to reference the old name for existing data (files, infra object names, etc) that need to remain accessible using the old name.


api/edgeproto/unmarshal_bc.go
```diff
@@ -0,0 +1,147 @@
...
+	"encoding/json"
+	strings "strings"
+
+	"github.com/edgexr/jsonparser"
+)
+
+// Support for unmarshaling older-object formatted JSON into current
+// objects for backwards compatibility. These functions shall be
+// used at the Master Controller to process data from old
+// CLI/UI clients that are sending the older format JSON.
```
levshvarts, line 27: Is this here specifically because you don't want to update UI just yet?

api/edgeproto/version.proto
```diff
@@ -29,13 +29,14 @@ enum VersionHash {
...
   //interim versions deleted
   HASH_611b28894b117c2aaa22c12adcd81f74 = 47; // FedAppInstKey
-  HASH_37dea30756fed2b0c0ecbc3e7b084855 = 48 [(protogen.upgrade_func) = "AppInstKeyName"];
+  HASH_37dea30756fed2b0c0ecbc3e7b084855 = 48; // AppInstKeyName - upgrade function removed
   HASH_1304c4ec69343ced28fd3ebc85f4a3a9 = 49; // Bad PlatformType upgrade that was a no-op
-  HASH_601fa4f6a8109f39e46adf1ea3b89197 = 50 [(protogen.upgrade_func) = "PlatformType"];
-  HASH_a61a29cd41f6b7459b05b6f7be6be4ce = 51 [(protogen.upgrade_func) = "CloudletAccessVars"];
+  HASH_601fa4f6a8109f39e46adf1ea3b89197 = 50; // PlatformType
+  HASH_a61a29cd41f6b7459b05b6f7be6be4ce = 51; // CloudletAccessVars
   HASH_c2d882033b0c14f28cece41cf4010060 = 52;
```
levshvarts, line 36: To avoid the confusion might be a good idea to just delete these hashes as I doubt we'll ever encounter them.

api/edgeproto/unmarshal_bc.go
gainsley: Exactly :)

api/edgeproto/version.proto
gainsley: True, I can do that.

pkg/controller/appinstclient_api_test.go
```diff
@@ -123,9 +125,13 @@ func TestAppInstClientApi(t *testing.T) {
 	assert.Equal(t, -1, count)
 	// Add a second Channel for AppInst1
 	ch12 := make(chan edgeproto.AppInstClient, qSize)
-	apis.appInstClientApi.SetRecvChan(ctx, &testutil.AppInstClientKeyData()[0], ch12)
+	wg := apis.appInstClientApi.SetRecvChan(ctx, &testutil.AppInstClientKeyData()[0], ch12)
```
levshvarts, line 128: that's a good catch.

pkg/controller/upgrade_funcs.go
levshvarts, line 1: Wow, git did a terrible job merging this file :)

pkg/controller/upgrade_funcs.go
```diff
@@ -540,171 +305,351 @@ func CloudletAccessVars(ctx context.Context, objStore objstore.KVStore, allApis
...
-	cloudletKeys, err := getDbObjectKeys(objStore, "Cloudlet")
+// InstanceKeysRegionScopedName deals with moving the CloudletKey out
+// of the AppInst/ClusterInst Key and onto the object body, changing
+// the unique key. Also, this requires the instance name to be unique
+// within the entire region instead of within a cloudlet scope.
+func InstanceKeysRegionScopedName(ctx context.Context, objStore objstore.KVStore, allApis *AllApis, sup *UpgradeSupport) error {
+	// Unfortunately the last upgrade func had a bug which failed to
+	// asign object IDs to AppInsts and ClusterInsts. We run the fixed
+	// version here again. Note that all upgrade functions must be idempotent.
+	err := UpgradeCrmOnEdge(ctx, objStore, allApis, sup)
```
levshvarts, line 316: Did you actually upgrade any setups already? If not, a fixed `UpgradeCrmOnEdge` would take care of things, but no harm in calling it again :)

levshvarts (approved):
Looks good. A lot of changes, but pretty straightforward. Thanks for cleaning up some old stuff too.

pkg/controller/upgrade_funcs.go
gainsley: Yeah, I noticed that...probably easier to just review the raw file. It in general does a really bad job if you remove a bunch of code and add a completely new function.

pkg/controller/upgrade_funcs.go
gainsley: I guess no setups are upgraded, although someone could technically pull up to a certain commit and upgrade on their own. I think it's safer to call it again.

gainsley:
Thanks Lev. I'm going through the corresponding edge-cloud-director changes, looks like I may need to make a few tweaks to this code in support, but just going to be some minor changes.

api/edgeproto/unmarshal_bc.go
gainsley: Ah, I just realized there's a flaw with this - I can deal with the request data, but without a client version, I can't send out the expected response data (like for AppInstShow). I think I will have to fix the UI.

PR #358, opened 2024-08-20
track etcd database version ID

gainsley:
This tracks the data model version ID in addition to the hash value in the etcd database. The model version ID is a monotonically increasing integer, and can be used to determine if an upgrade has been upgrade or not, regardless of whether the current software is still tracking the upgrade associated with it.

The intent here is to be able to tag objects with a db model version ID to guarantee that upgrade functions are idempotent. So far upgrade functions guarantee idempotency by looking at the current data in the object. This does not work for new non-nil fields that need to be set on upgrade, but may be cleared by the user later on. Consider the "CrmOnEdge" bool. On upgrade we need to set it "true" because all existing cloudlets use Crms on Edge. However, new cloudlets may not use crm on edge, so may set it to false. Therefore there is no way to tell from a false CrmOnEdge value if the cloudlet needs to be upgraded or not. As an aside, the CrmOnEdge upgrade function guarantees idempotency by looking at the "ObjId" field which does not have the same issue, and is part of the same upgrade.

The intent here then is to add a "dbModelVersion int32" field to an object, like Cloudlet, and compare that to the ID for the running upgrade function to tell if the object needs to be upgraded or not. This will be used for the Zone upgrade feature which adds a string field to the Cloudlet.

### DIFF
--- a/api/edgeproto/version.pb.go
+++ b/api/edgeproto/version.pb.go
@@ -250,8 +250,15 @@ var VersionHashCommonPrefix = "Hash"
 // TrustPolicyExceptionKey
 // VMPoolKey
 // VirtualClusterInstKeyV1
-var versionHashString = "eac56710c013d954db31eeb306b514a4"
 
-func GetDataModelVersion() string {
-	return versionHashString
+type DataModelVersion struct {
+	Hash string
+	ID   int32
+}
+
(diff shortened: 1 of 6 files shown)


levshvarts (approved):
Yep, that makes sense. Changes look good.

gainsley:
Thanks Lev 👍

PR #360, opened 2024-08-28
clusterrefs upgrade fix

gainsley:
The AppInstKeyName upgrade function failed to upgrade ClusterRefs properly. As a result the database is now in a state of old and new clusterRefs, the old ones which still refer to AppInsts by AppKey+CloudletKey.

This change fixes an issue with the InstanceKeysRegionScopedName function hitting errors on these old formats, and also adds another upgrade function to fix the old clusterRefs.

The additional upgrade function fixes clusterRefs by rebuilding them from existing AppInsts.


levshvarts (approved):
Looks good. Thanks for fixing this.

Issue #249, opened 2023-12-19
Openstack: save rootLB object name on Cloudlet


gainsley:
Openstack generates the shared rootLB name from pkg/platform/common/vmlayer/lb.go:GetRootLBName.

Unfortunately this is generated dynamically, so if for instance the AppDNSRoot changes due to changing the domain name, now the name used to refer to the heat stack in Openstack changes, and the CRM thinks it needs to create a new rootLB with the new name, ignoring the old rootLB with it's existing configurations/proxies.

At cloudlet creation time we should decide upon the shared rootLB name and record it on the Cloudlet object. In fact we do have a cloudlet.RootLBFQDN name. The problem there is we need to separate URIs from object names. Because when we migrate domain names, we'll eventually want to migrate the URIs to the new domain as well, but the object names should never need to change.

levshvarts:
TODO - need to handle `crmOffEdge` case still

PR #361, opened 2024-08-29
Cloudlet DNS migration 

levshvarts:
### Issues Fixed

* https://github.com/edgexr/edge-cloud-platform/issues/249

### Description

Added a new api `changedns`(`/change/cloudlet/dns`)
This will update the following objects Cloudlet.rootLBFQDN, ClusterInst.Fqdn, AppInst.Uri
Also added platform interfaces to update DNS entries specifically - `ChangeAppInstDNS`, `ChangeClusterInstDNS`, `ChangeClusterInstDNS`

Internally we rely on already implemented platform code to register DNS names as well as replace the certs.


pkg/platform/common/vmlayer/appinst.go
```diff
@@ -914,6 +917,33 @@ func (v *VMPlatform) UpdateAppInst(ctx context.Context, clusterInst *edgeproto.C
...
+	err = v.SetupRootLB(ctx, appInst.Uri, appInst.Uri, &appInst.CloudletKey, &pp, rootLBDetail, appInst.EnableIpv6, updateCallback)
+	if err != nil {
+		return err
+	}
+
+	// Now delete old DNS entry
+	updateCallback(edgeproto.UpdateTask, "Deleting old DNS")
+	if err := v.VMProperties.CommonPf.DeleteDNSRecords(ctx, OldURI); err != nil {
+		log.SpanLog(ctx, log.DebugLevelInfra, "failed to delete DNS record", "uri", OldURI, "err", err)
+	}
```
gainsley, line 942: One thing that's going to be an issue when you add the GRPC API for CCRM is that the CCRM does not have the "old" state of the object available. That's only available here in the CRM because notify goes through a cache and can pull out the old state from the cache. In CCRM, we'd typically just send the new AppInst object as the GRPC Request object. Since you've already written the new state to etcd, CCRM also can only look up the new state from Etcd.

One way to deal with this is to make a custom object (instead of just AppInst) to pass via the GRPC call that can also carry the old URI. But that makes the API messy, it'd be better to keep the API as simple CRUD APIs between Controller and CCRM as much as possible.

Alternatively you could add an annotation to the AppInst which is the old URI. That may be useful information to have. It would also allow us to recover if CRM crashed before it was able to delete the old URI (because you've already written the new URI to etcd at that point, once the update process at the Controller times out the old URI will be lost forever). It's not really that big of a deal here if we fail to clean up the old DNS entry, of course, but it's worth considering.

A third option to actually do the DNS entry delete from the Controller once its finished waiting for the CRM to finish updating. I dislike this option though because it's the platform-specific code that's managing DNS entries, and the Controller doesn't really know whether the platform actually needs to delete a DNS entry or not. I think it's cleaner to keep that responsibility in the platform-specific code.

gainsley (commented):
Thanks Lev for adding this!

pkg/platform/common/vmlayer/appinst.go
levshvarts: This is a good point - I did notice CCRM apis just taking the object. Let me add annotations in order to have the record of old uri.

pkg/controller/appinst_api.go
```diff
@@ -2372,6 +2372,44 @@ func (s *AppInstApi) RecordAppInstEvent(ctx context.Context, appInst *edgeproto.
 	services.events.AddMetric(&metric)
 }
 
+func (s *AppInstApi) updateURI(key *edgeproto.AppInstKey, cloudlet *edgeproto.Cloudlet, inCb edgeproto.AppInstApi_UpdateAppInstServer) (reterr error) {
+	ctx := inCb.Context()
+	cctx := DefCallContext()
+
+	log.SpanLog(ctx, log.DebugLevelApi, "updateRootLbFQDN", "cluster", key, "cloudlet", cloudlet)
+	streamCb, cb := s.all.streamObjApi.newStream(ctx, cctx, key.StreamKey(), inCb)
```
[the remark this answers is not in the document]
levshvarts: I guess not - I can just pass this stream along the function calls.

pkg/controller/appinst_api.go
```diff
@@ -2372,6 +2372,44 @@ func (s *AppInstApi) RecordAppInstEvent(ctx context.Context, appInst *edgeproto.
...
+	reqCtx, reqCancel := context.WithTimeout(cb.Context(), s.all.settingsApi.Get().UpdateAppInstTimeout.TimeDuration())
+	defer reqCancel()
+
+	successMsg := ""
+	if !strings.Contains(key.Name, cloudcommon.MEXPrometheusAppName) && !strings.Contains(key.Name, cloudcommon.NFSAutoProvisionAppName) {
+		successMsg = fmt.Sprintf("AppInst %s updated successfully", key.Name)
+	}
+	return edgeproto.WaitForAppInstInfo(reqCtx, key, s.store, edgeproto.TrackedState_READY,
+		UpdateAppInstTransitions, edgeproto.TrackedState_UPDATE_ERROR,
+		successMsg, cb.Send, edgeproto.WithCrmMsgCh(sendObj.crmMsgCh))
```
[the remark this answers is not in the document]
levshvarts: You are right. So I think the way it is working is that CRM gets notified of objects change and does the work without really needing to report back to the controller.  So I check the logs in the crm and see all the bits activating, just don't see the fact that it's not really doing anything(And I do notice that messages don't get streamed back, as the stream is already closed.)

pkg/controller/cloudlet_api.go
```diff
@@ -2625,3 +2625,101 @@ func (s *CloudletApi) updateDefaultMultiTenantClusterWorker(ctx context.Context,
...
+	defer reqCancel()
+
+	successMsg := fmt.Sprintf("Cloudlet %s FQDN updated successfully", key.Name)
+	if cloudlet.CrmOnEdge {
+		// Wait for cloudlet to finish updating DNS entries
+		err = edgeproto.WaitForCloudletInfo(
+			reqCtx, key, s.all.cloudletInfoApi.store,
+			dme.CloudletState_CLOUDLET_STATE_READY,
+			UpdateCloudletTransitions, dme.CloudletState_CLOUDLET_STATE_ERRORS,
+			successMsg, cb.Send,
```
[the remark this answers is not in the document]
levshvarts: See the comment above - it's working, but not really waiting for crm to finish. Let me rewrite some parts here.

pkg/platform/common/vmlayer/appinst.go
```diff
@@ -217,10 +217,13 @@ func (v *VMPlatform) setupVMAppRootLBAndNode(ctx context.Context, clusterInst *e
...
-	err = v.proxyCerts.SetupTLSCerts(ctx, orchVals.lbName, orchVals.lbName, client)
-	if err != nil {
-		return err
-	}
+
+	// Why? This is done in SetupRootLB
+	//	err = v.proxyCerts.SetupTLSCerts(ctx, orchVals.lbName, orchVals.lbName, client)
+	//	if err != nil {
+	//		return err
+	//	}
```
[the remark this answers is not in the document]
levshvarts: I left the comment in for now, as I was going to verify this first. Will remove after double-checking.

pkg/crm/crm_appinst.go
```diff
@@ -68,6 +68,14 @@ func (s *CRMData) appInstChanged(ctx context.Context, old *edgeproto.AppInst, ne
...
 			} else {
 				new.Fields = old.GetDiffFields(new).Fields()
+				// Special case for dns update - only possible if appinst exists
+				fmap := edgeproto.MakeFieldMap(new.Fields)
+				if !fmap.Has(edgeproto.AppInstFieldState) {
+					if fmap.Has(edgeproto.AppInstFieldUri) {
+						_ = s.AppInstDNSChanged(ctx, s.cloudletKey, old, new, responseSender)
+						return
+					}
+				}
```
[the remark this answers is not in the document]
levshvarts: I was planning to add a separate DNS Change API for the CCRM, since this is so specialized and we might not want to go through the same logic as a generalized update. Would you rather use the existing update api from CCRM?
You other comment about Adding Annotation for the previous dns name seem to suggest you'd prefer reusing existing update api to CCRM. Am I getting it right?

pkg/crm/crm_appinst.go
```diff
@@ -68,6 +69,11 @@ func (s *CRMData) appInstChanged(ctx context.Context, old *edgeproto.AppInst, ne
 				new.Fields = edgeproto.AppInstAllFields
 			} else {
 				new.Fields = old.GetDiffFields(new).Fields()
+				oldDNS, ok := new.Annotations[cloudcommon.AnnotationPreviousDNSName]
+				if ok && oldDNS == old.Uri {
+					_ = s.AppInstDNSChanged(ctx, s.cloudletKey, old, new, responseSender)
```
gainsley, line 74: This still depends on old so this logic won't translate to the CCRM, I think it would be more straightforward to just check
```
if old.Uri != new.Uri {
```
In the CCRM, we don't have to worry about multiple updates via notify that could erroneously trigger another DNS update. In the CCRM if we get a call to Update the DNS, it's because the Controller decided it needed to be done. We just need the old URI to be able to delete the old DNS entry, not to determine whether or not to perform the action.

gainsley (commented):
Thanks for making those changes!

pkg/controller/cloudlet_api.go
```diff
@@ -2625,3 +2625,114 @@ func (s *CloudletApi) updateDefaultMultiTenantClusterWorker(ctx context.Context,
...
+
+	// Step 1 - update rootLb fqdn
+	// Does our current DNS app root name match what the clouldet alredy have?
+	// TODO - why does the streaming not work for cloudlet crm update
+	if strings.HasSuffix(cloudlet.RootLbFqdn, *appDNSRoot) {
+		cb.Send(&edgeproto.Result{Message: "Cloudlet rootLB is already set correctly"})
+		log.SpanLog(ctx, log.DebugLevelApi, "Current cloudlet fqdn already contains appDNSRoot suffix - nothing to do")
+	} else {
+		log.SpanLog(ctx, log.DebugLevelApi, "Update rootLB fqdn", "old", cloudlet.RootLbFqdn, "new", getCloudletRootLBFQDN(&cloudlet))
+		cb.Send(&edgeproto.Result{Message: "Updating Cloudlet RootLb FQDN"})
```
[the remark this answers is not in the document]
levshvarts: I thought about it, and I'd rather have the cloudlet in maintenance for the duration of the api - this way it eliminates possible race-conditions with appInsts, clusterInsts being modified.

pkg/crm/crm_appinst.go
levshvarts: Yeah, I didn't make this change yet - I pushed some intermediate changeset, which didn't address your comments about moving the logic into controller-data.go yet. Will fix.

pkg/controller/appinst_api.go
```diff
@@ -2372,6 +2370,54 @@ func (s *AppInstApi) RecordAppInstEvent(ctx context.Context, appInst *edgeproto.
...
+			return nil
+		}
+		if appInst.Uri == "" {
+			needUpdate = false
+		}
+		// Store previous URI, so we can update this with CCRM
+		appInst.AddAnnotation(cloudcommon.AnnotationPreviousDNSName, appInst.Uri)
+		appInst.Uri = getAppInstFQDN(&appInst, cloudlet)
+		appInst.UpdatedAt = dme.TimeToTimestamp(time.Now())
+		appInst.State = edgeproto.TrackedState_UPDATE_REQUESTED
```
gainsley, line 2395: Btw just realized, this needs to check the current appInst.State to make sure it's in Ready state, or if the user specified an appropriate crmIgnore state. You should be able to look at UpdateClusterInst for the logic.

levshvarts: Hm, you are right - it will set the state to ready at the end of this. Also will need to check the state for clusterInsts.
For the simplicity I'd prefer to either skip the non-ready objects, or rather error our when we encounter the first one. Which way do you think is better?

With respect to crmIgnore - this api doesn't actually have it, since it takes a key for the cloudlet rather than the whole clouldet object.

gainsley: I think we skip the non-ready objects. I'm assuming we can rerun the command again and fix all the objects we missed the first time around.

Ok, I think we don't need crm ignore, as they should be able to use the normal Update API to fix any bad state. Just need to check the current state so we don't overwrite another change in progress.

pkg/crm/crm_appinst.go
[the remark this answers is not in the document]
levshvarts: Moved the logic to controller-data.go.

pkg/controller/cloudlet_api_test.go
```diff
@@ -1125,3 +1127,114 @@ func testCloudletEdgeboxOnly(t *testing.T, ctx context.Context, cloudlet edgepro
...
+			apis.cloudletInfoApi.cache.Update(ctx, &info, 0)
+		}
+	})
+
+	defer cancel()
+
+	cloudlet.MaintenanceState = dme.MaintenanceState_MAINTENANCE_START
+	cloudlet.Fields = append(cloudlet.Fields, edgeproto.CloudletFieldMaintenanceState)
+	err = apis.cloudletApi.UpdateCloudlet(&cloudlet, testutil.NewCudStreamoutCloudlet(ctx))
+	require.Nil(t, err, fmt.Sprintf("update cloudlet maintenance state"))
```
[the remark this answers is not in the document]
levshvarts: I couldn't use `cloudletApi.store.Put()` directly, since it does object validation and wants Maintenance_start instead of Under_maintenance, but I was able to just use `cloudetApi.setMaintenanceState()` here.

pkg/controller/clusterinst_api.go
```diff
@@ -2059,3 +2059,51 @@ func (s *ClusterInstApi) deleteCloudletSingularCluster(stm concurrency.STM, key
...
+func (s *ClusterInstApi) updateRootLbFQDN(key *edgeproto.ClusterKey, cloudlet *edgeproto.Cloudlet, inCb edgeproto.ClusterInstApi_UpdateClusterInstServer) (reterr error) {
+	ctx := inCb.Context()
+	cctx := DefCallContext()
+
+	log.SpanLog(ctx, log.DebugLevelApi, "updateRootLbFQDN", "cluster", key, "cloudlet", cloudlet)
+
+	clusterKey := key
+	streamCb, cb := s.all.streamObjApi.newStream(ctx, cctx, clusterKey.StreamKey(), inCb)
+
+	modRev, _ := s.sync.ApplySTMWaitRev(ctx, func(stm concurrency.STM) error {
```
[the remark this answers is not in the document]
levshvarts: fixed.

pkg/controller/clusterinst_api.go
```diff
@@ -2059,3 +2059,51 @@ func (s *ClusterInstApi) deleteCloudletSingularCluster(stm concurrency.STM, key
...
+
+		// save old dns fqdn, so it can be updated in CCRM
+		clusterInst.AddAnnotation(cloudcommon.AnnotationPreviousDNSName, clusterInst.Fqdn)
+		clusterInst.Fqdn = getClusterInstFQDN(&clusterInst, cloudlet)
+		clusterInst.UpdatedAt = dme.TimeToTimestamp(time.Now())
+		clusterInst.State = edgeproto.TrackedState_UPDATE_REQUESTED
+		s.store.STMPut(stm, &clusterInst)
+		return nil
+	})
+
```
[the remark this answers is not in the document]
levshvarts: fixed.

pkg/crm/crm_appinst.go
levshvarts: Updated the code to handle the transitions in the `controller-data.go` instead so CCRM can use the same logic.

pkg/crmutil/controller-data.go
```diff
@@ -200,12 +200,40 @@ type NeedsUpdate struct {
...
 	var err error
 
 	fmap := edgeproto.MakeFieldMap(new.Fields)
 
-	if !fmap.Has(edgeproto.ClusterInstFieldState) {
+	dnsUpdate := false
+	oldFQDN, ok := new.Annotations[cloudcommon.AnnotationPreviousDNSName]
+	if fmap.Has(edgeproto.ClusterInstFieldFqdn) && ok && oldFQDN != new.Fqdn {
+		dnsUpdate = true
+	}
```
gainsley, line 234: You should just check if the field flag is set. The oldFQDN in the annotation is always going to be different than the current FQDN.

pkg/crmutil/controller-data.go
```diff
@@ -336,12 +370,41 @@ func (cd *CRMHandler) CloudletHasTrustPolicy(ctx context.Context, cloudletKey *e
...
 
-	var err error
 	fmap := edgeproto.MakeFieldMap(new.Fields)
-	if !fmap.Has(edgeproto.AppInstFieldState) {
+	dnsUpdate := false
+	oldURI, ok := new.Annotations[cloudcommon.AnnotationPreviousDNSName]
+	if fmap.Has(edgeproto.AppInstFieldUri) && ok && oldURI != new.Uri {
+		dnsUpdate = true
+	}
+
```
gainsley, line 406: Same thing here, just check for the field flag, since the annotation is always different from the new.Uri.

pkg/crmutil/controller-data.go
```diff
@@ -544,20 +613,64 @@ func (cd *CRMHandler) appInstInfoRuntime(ctx context.Context, sender edgeproto.A
...
 		return err
 	}
-	fmap := edgeproto.MakeFieldMap(new.Fields)
 
+	// Special case for dns update
+	oldDNS, ok := new.Annotations[cloudcommon.AnnotationPreviousDNSName]
+	if ok && oldDNS != new.RootLbFqdn {
+		_ = cd.CloudletDNSChanged(ctx, pf, oldDNS, new, sender)
+		return
+	}
```
gainsley, line 671: This should check field flag instead. The field flag is computed from the old state in the cache vs the new received state, so accurately detects a change. The oldDNS annotation is always going to be different from the current RootLbFqdn when this function receives another unrelated update at a later time.

pkg/controller/cloudlet_api.go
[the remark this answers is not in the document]
levshvarts: Sounds good. The scenario I was thinking about is that since we are first collecting clusterInsts and only after that updating them one-by-one, there is a moment between these two actions that a new cluster could be created.  But in any case - I'll keep maintenance requirement for the whole api work.

pkg/crmutil/controller-data.go
levshvarts: that's fair - `oldFQDN != new.Fqdn` is a noop check. I'll keep the check for the existence of annotation though.

pkg/crmutil/controller-data.go
levshvarts: Same as above - I keep the check that Annotations actually exist, otherwise this gets triggered on the create.

pkg/controller/appinst_api.go
```diff
@@ -1176,6 +1179,7 @@ func (s *AppInstApi) createAppInstInternal(cctx *CallContext, in *edgeproto.AppI
 			if cloudletCompatibilityVersion >= cloudcommon.CRMCompatibilitySharedRootLBFQDN {
 				// no external access to AppInst, no need for URI
 				in.Uri = ""
+				in.StaticUri = ""
```
gainsley, line 1182: All these instances of `in.StaticUri = in.Uri` can be consolidated to one instance right before calling`s.store.STMPut(stm, in)`.

gainsley (approved):
Looks good!

pkg/controller/appinst_api.go
```diff
@@ -409,6 +409,37 @@ func removeAppInstFromRefs(appInstKey *edgeproto.AppInstKey, appInstRefs *[]edge
...
+		return ""
+	}
+
+	// Default to dedicated IP
+	uri := getAppInstFQDN(appInst, cloudlet)
+
+	// uri is specific to appinst if it has dedicated IP
+	if appInst.DedicatedIp {
+		return uri
+	}
```
gainsley, line 427: You can just change line 426 to
```
return getAppInstFQDN(appInst, cloudlet)
```
and avoid computing uri on line 422 when it's not needed.

gainsley (approved):
looks good

pkg/controller/appinst_api.go
levshvarts: I updated the logic, and got rid of the `uri` local var

PR #362, opened 2024-08-29
upgrade fixes again

gainsley:
The previous upgrade function had two problems:
- For AppInsts, we did not update the ClusterInst key with the new name if the ClusterInst name was changed.
- For reservable ClusterInsts, the name format generated by the upgrade function was incorrect. It was just appending a number to avoid conflicts, however for reservable ClusterInsts the system wants to append the cloudlet hash. This supports the previous method of using IDs to generate unique names.

This upgrade function must be able to upgrade old data, broken data from the previous upgrade, and new data (no-op).

Unfortunately because I needed to swap the order of the ClusterInsts and AppInsts processing in the upgrade function, the diff is a bit ugly. There's really not that many changes.


levshvarts (approved):
Looks good.

PR #363, opened 2024-08-30
upgrade func not run

gainsley:
With the change to use IDs instead of Hashes to select upgrade functions, a bug was introduced where upgrade functions were not being run because the for loop was looking for a "nextVers" ID in the map that didn't exist. This was happening in `upgradeToLatest`. This happened because a VersionV2 was not found in the db, and we read the hash from Version but set the ID to 0. In this case, `nextVers` became 1, for which there was no associated entries in any of the maps, causing the for loop to break out immediately.

I've changed the for loop to walk the list of possible upgrade functions, rather than looping over possible ID values.

I also added a unit test to verify the correct upgrade functions will be run. This required refactoring some of the existing code.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -265,22 +265,8 @@ func startServices() error {
 
 	// We might need to upgrade the stored objects
 	if !*skipVersionCheck {
-		// First off - check version of the objectStore we are running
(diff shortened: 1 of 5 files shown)


levshvarts (approved):
Looks good. It's a bit more prescriptive about the sequence. Thanks for fixing this.

PR #364, opened 2024-09-03
Delete consistency check after upgrade test

gainsley:
A recent upgrade function had a bug that caused a failure to delete existing objects after upgrade because a reference was missing. This change adds a delete test after the upgrade test to check that deletes will work after the upgrade is done. The intended use is to test new upgrade functions by taking a dump of an existing deployment and feeding it through the upgrade test. The deletes mimic what a user may do with existing objects.

We may want to consider more consistency checks later as well, in particular verifying refs objects data.

### DIFF
--- a/pkg/controller/upgrade_test.go
+++ b/pkg/controller/upgrade_test.go
@@ -46,9 +46,7 @@ var upgradeVaultTestFilePostSuffix = "_post.vault"
 var upgradeVaultTestFileExpectedSuffix = "_expected.vault"
 
 // Walk testutils data and populate objStore
-func buildDbFromTestData(objStore objstore.KVStore, funcName string) error {
-	ctx := context.Background()
-
+func buildDbFromTestData(ctx context.Context, objStore objstore.KVStore, funcName string) error {
 	filename := upgradeTestFileLocation + "/" + funcName + upgradeTestFilePreSuffix
 	file, err := os.Open(filename)
 	if err != nil {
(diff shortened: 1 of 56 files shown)


pkg/controller/upgrade_test.go
```diff
@@ -440,6 +441,33 @@ func TestAllUpgradeFuncs(t *testing.T) {
...
+		log.SpanLog(ctx, log.DebugLevelInfo, "appinstance", "ai", allData.AppInstances[ii].Key.GetKeyString())
+		allData.AppInstances[ii].CrmOverride = edgeproto.CRMOverride_IGNORE_CRM_AND_TRANSIENT_STATE
+	}
+	for ii := range allData.ClusterInsts {
+		allData.ClusterInsts[ii].CrmOverride = edgeproto.CRMOverride_IGNORE_CRM_AND_TRANSIENT_STATE
+	}
+	for ii := range allData.Cloudlets {
+		allData.Cloudlets[ii].CrmOverride = edgeproto.CRMOverride_IGNORE_CRM_AND_TRANSIENT_STATE
+	}
+	for ii := range allData.VmPools {
```
levshvarts, line 465: Why VmPools specifically?

levshvarts: Never mind - I see it just sets the CRMOverride state.

levshvarts (approved):
Looks good. I think deletion of the objects is sufficient for now. It will at least allow for the broken objects to be cleaned up by the user.

PR #366, opened 2024-09-09
add zones feature

### LINKED ISSUE #158: Add Zone objects
Zones are groups of cloudlets that have a similar geographical location and network performance. The user can choose to deploy to a Zone, and the platform will decide which cloudlet in the Zone to deploy to. The definition comes from the OPG definitions for operator platforms and federation APIs.

Currently, for federation, the operator must create a zone that corresponds to only one cloudlet. These zones are defined only for federation, and not for local users to use. There is no code for automating placement within a zone of more than one cloudlet.

gainsley:
This adds the Zones feature as described by [OPG spec](https://www.gsma.com/solutions-and-impact/technologies/networks/wp-content/uploads/2023/07/OPG.02-v5.0-Operator-Platform-Requirements-and-Architecture.pdf).

A zone groups a set of cloudlets together that share the same failure domain. Different zones should have decoupled resource dependencies, meaning a infrastructure/cloudlet failure in one Zone should never cause a failure in another Zone. App providers deploy their applications to a zone, and the system decides which cloudlet in the zone. Application providers are not allowed to know about cloudlets or which cloudlet their application is deployed on.

Because clusters and app instances are actually located on cloudlets, we implement zones as more of an aliasing mechanism than a strict object. Therefore the Zone object does not contain a list of cloudlets in the zone, but rather each cloudlet tracks which zone it belongs to. Additionally, the "zone" for an app/cluster instance is always derived from the zone defined on the cloudlet. Although we track the zone key on app/cluster instances, that value should be considered a cached value, with the source of truth being the zone defined on the cloudlet which is referenced by the instance's cloudletKey.

Cloudlets and Instances may have an empty ZoneKey, meaning they belong to no zone. This facilitates off-lining Zones and moving cloudlets and their associated app/cluster instances between Zones. Even with an empty zone key, users can still update/delete existing cloudlets/instances.

CloudletPools are now ZonePools. Although similar in nature, Zones define a grouping of cloudlets based on resource dependencies for redundancy. ZonePools define a grouping of cloudlets (now zones) based on segregation between organizations for accessibility (for example a Production ZonePool and a Staging ZonePool). Also, a single cloudlet may only belong to one Zone, but a Zone may belong to 0 or more ZonePools.

AutoProvPolices now are based on a collection of Zones instead of Cloudlets, because these are developer-facing policies, and developers do not have visibility of cloudlets.

Metrics now also need to be tagged with both Cloudlet keys and Zone keys. We use the MC to determine from RBAC if the user is allowed to see the cloudlet key tags.

A general note on the new deployment algorithms for Create ClusterInst/AppInst: All the cloudlet-specific checks have now been refactored into a potentialCloudlets check, which assumes a zone may consist of a heterogeneous set of cloudlets, and which filters out cloudlets in the Zone that cannot be used for the specific instance. In the AppInst case, we then search from the potentialCloudlets for potentialClusters, which are pre-existing clusters that can be used to deploy the application. The potentialClusters search is just code refactored from the previous code that looked at default/MT/reservable clusters.

Fixes #158.


pkg/controller/appinst_zone.go
```diff
@@ -0,0 +1,323 @@
...
+			skipReasons.add(skipReason)
+			continue
+		}
+		potentialCloudlets = append(potentialCloudlets, pc)
+	}
+	if len(potentialCloudlets) == 0 {
+		reasonsStr := skipReasons.String()
+		if in.ClusterKey.Name != "" {
+			if reasonsStr == "" {
+				return nil, fmt.Errorf("cannot deploy to cluster %s", in.ClusterKey.Name)
```
levshvarts, line 95: What case would that be? At least we probably need to log this case, as this seems like a bug somewhere.

pkg/controller/appinst_zone_test.go
```diff
@@ -0,0 +1,216 @@
...
+	pclos, err = apis.appInstApi.getPotentialCloudlets(ctx, cctx, ai, app)
+	require.Nil(t, err)
+	verifyCloudlets(pclos, 3, 1, 2, 0)
+	pclus, err = apis.appInstApi.getPotentialClusters(ctx, cctx, ai, app, pclos)
+	require.Nil(t, err)
+	require.Equal(t, 1, len(pclus))
+	require.Equal(t, reservK.Key, pclus[0].existingCluster)
+
+	// Docker app with serverless config
+	// should find only reservD cluster
```
levshvarts, line 189: If it finds the same cluster as the test case on line 166, doesn't it need to be checked again?

pkg/controller/cloudlet_api.go
```diff
@@ -2079,6 +2092,19 @@ func (s *CloudletApi) UsesResTagTable(key *edgeproto.ResTagTableKey) *edgeproto.
 	return nil
 }
 
+func (s *CloudletApi) UsesZone(zoneKey *edgeproto.ZoneKey) []string {
```
levshvarts, line 2095: I would change this to `cloudletsInZone`, or something, otherwise this sounds like it should return a bool.

pkg/controller/cloudlet_api.go
```diff
@@ -2626,6 +2653,133 @@ func (s *CloudletApi) updateDefaultMultiTenantClusterWorker(ctx context.Context,
...
+			} else {
+				var lat, long float64
+				for _, loc := range locs {
+					lat += loc.Latitude
+					long += loc.Longitude
+				}
+				lat /= float64(len(locs))
+				long /= float64(len(locs))
+				zone.Location.Latitude = lat
+				zone.Location.Longitude = long
```
levshvarts, line 2772: Oh man, zone location could really be random with these calculations. Maybe specifying the location when creating a zone makes more sense? What do you think?

pkg/controller/platformfeatures_api.go
```diff
@@ -101,6 +102,55 @@ func (s *PlatformFeaturesApi) DeletePlatformFeatures(ctx context.Context, in *ed
 	return s.store.Delete(ctx, in, s.sync.SyncWait)
 }
 
+func (s *PlatformFeaturesApi) ShowPlatformFeaturesForZone(key *edgeproto.ZoneKey, cb edgeproto.PlatformFeaturesApi_ShowPlatformFeaturesForZoneServer) error {
```
levshvarts, line 105: It looks like this returns a superset of features for all the cloudlets in this zone. For my understanding is the intent to return all the supported features and then use `potentialCloudlet` methods to get the cloudlet that support the features if one is required for an app, right?

pkg/controller/potential_cloudlet.go
```diff
@@ -0,0 +1,142 @@
...
+	if iscore == jscore {
+		return a[i].cloudlet.Key.GetKeyString() < a[j].cloudlet.Key.GetKeyString()
+	}
+	return iscore > jscore
+}
+
+const (
+	ResourceWeightVCPU uint64 = 1000
+	ResourceWeightRAM  uint64 = 1
+)
```
levshvarts, line 142: `Const` at the end of the file is a bit weird :)

pkg/controller/upgrade_funcs.go
```diff
@@ -879,3 +881,280 @@ func InstanceKeysRegionScopedName(ctx context.Context, objStore objstore.KVStore
...
+				return nil
+			}
+			cloudlet := edgeproto.Cloudlet{}
+			if err2 := unmarshalUpgradeObj(ctx, cloudletStr, &cloudlet); err2 != nil {
+				return err2
+			}
+			if cloudlet.DbModelId >= dbModelID {
+				// already upgraded
+				return nil
+			}
```
levshvarts, line 949: Are you thinking we might hit an issue where we'd need to run this upgrade function multiple times?

levshvarts (commented):
Still reviewing, but here are some of my comments.

api/edgeproto/appinst.proto
```diff
@@ -224,6 +226,8 @@ message AppInst {
   string obj_id = 49 [(protogen.backend) = true, (protogen.hidetag) = "nocmp"];
   // Annotations
   map<string, string> annotations = 52;
+  // database version model ID
```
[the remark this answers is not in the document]
gainsley: It's in case the upgrade function is run again, or let's say there are two instances of the regional controller and they get restarted one after the other, then the second controller will run the upgrade function that the first controller already ran. This lets us know the object was already upgraded.

api/edgeproto/objs.go
```diff
@@ -1128,8 +1148,8 @@ func CmpSortSlices() []cmp.Option {
 	opts = append(opts, cmpopts.SortSlices(CmpSortAppInstInfo))
 	opts = append(opts, cmpopts.SortSlices(CmpSortClusterInstInfo))
 	opts = append(opts, cmpopts.SortSlices(CmpSortNode))
-	opts = append(opts, cmpopts.SortSlices(CmpSortCloudletPool))
-	opts = append(opts, cmpopts.SortSlices(CmpSortCloudletPoolMember))
+	opts = append(opts, cmpopts.SortSlices(CmpSortZonePool))
```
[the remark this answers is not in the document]
gainsley: Yeah we do, thanks, I'll add that.

api/edgeproto/objs.go
```diff
@@ -455,6 +458,23 @@ func (s *Cloudlet) Validate(fmap objstore.FieldMap) error {
...
+	return nil
+}
+
+func (s *Zone) Validate(fmap objstore.FieldMap) error {
+	if err := s.GetKey().ValidateKey(); err != nil {
+		return err
+	}
+	return nil
+}
+
```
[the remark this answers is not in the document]
gainsley: I think if it's really boilerplate it would make sense to move it into the generator code. The validatekey function we could probably do by adding annotations to each key field for the validator function to use. Not sure if we could auto-gen the object validate function though.

pkg/controller/appinst_api.go
```diff
@@ -529,37 +526,10 @@ func (s *AppInstApi) createAppInstInternal(cctx *CallContext, in *edgeproto.AppI
...
-		if in.EnableIpv6 && !cloudletFeatures.SupportsIpv6 {
-			return fmt.Errorf("cloudlet platform does not support IPv6")
-		}
-		if !cloudcommon.IsClusterInstReqd(&app) && !in.EnableIpv6 {
-			// VM Apps default to the platform setting
-			in.EnableIpv6 = cloudletFeatures.SupportsIpv6
+		if err := s.all.autoProvPolicyApi.appInstCheck(ctx, stm, cloudcommon.Create, &app, in); err != nil {
+			return err
 		}
+		// make sure cloudlet exists so we don't create refs for missing cloudlet
```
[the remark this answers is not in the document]
gainsley: yup I'll remove it

pkg/controller/appinst_api.go
```diff
@@ -981,10 +871,8 @@ func (s *AppInstApi) createAppInstInternal(cctx *CallContext, in *edgeproto.AppI
 		clusterInst.Key = clusterKey
 		clusterInst.CloudletKey = in.CloudletKey
 		clusterInst.Auto = true
-		if autoClusterType == ReservableAutoCluster {
```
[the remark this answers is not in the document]
gainsley: Yes, originally they were non-reservable, and they would just get deleted when the AppInstance got deleted.

pkg/controller/appinst_zone.go
gainsley: Yeah I agree this is confusing. The only case that returns an empty reason is when they don't have permissions to deploy to a multi-tenant cluster. I don't want to display that reason if they're just searching for some cluster to use, but if they did specify that specific cluster, then we should send back the reason. I'll change it so the reason returned from the validatePotentialCloudlet() function will take into account whether they specified a specific cluster or not.

pkg/controller/appinst_zone_test.go
gainsley: Hm, not sure what you mean. Serverless config is really only for Kubernetes, so results should be the same whether or not docker app specifies allowServerless, that's what this is testing.

pkg/controller/cloudlet_api.go
gainsley: Changed to CloudletsUsingZone()

pkg/controller/cloudlet_api.go
gainsley: That's a fair point. I think it's probably ok for now, as any point within the "zone" area is as accurate as any other for the location. But I agree we may want to let the user specify the location at some point.

pkg/controller/potential_cloudlet.go
gainsley: moved it :)

api/edgeproto/autoprovpolicy.proto
```diff
@@ -62,31 +62,23 @@ message AutoProvPolicy {
...
   option (protogen.alias) = "name=Key.Name,apporg=Key.Organization";
   option (protogen.uses_org) = "key=Organization";
-  option (protogen.generate_lookup_by_sublist) = "CloudletKey";
-  option (protogen.noconfig) = "Cloudlets:#.Loc,DeletePrepare";
+  option (protogen.generate_lookup_by_sublist) = "ZoneKey";
+  option (protogen.noconfig) = "DeletePrepare";
 }
 
-// AutoProvCloudlet stores the potential cloudlet and location for DME lookup
-message AutoProvCloudlet {
```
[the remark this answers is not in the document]
levshvarts: Yep, found it.

api/edgeproto/objs.go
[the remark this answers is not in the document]
levshvarts: Yeah, it's either creating a validation function, or adding annotations in the proto file. It's a tradeoff I agree. Let me think about what possible things we can autogen, but I see how validation function is not as clear-cut  as I thought.

pkg/controller/cloudlet_api.go
levshvarts: Thanks!

pkg/controller/appinst_zone_test.go
levshvarts: Ah I see ok - didn't understand whether this is the test case :)

levshvarts (approved):
Looks good. Thanks for a massive refactor.

pkg/controller/appinst_zone_test.go
gainsley: let me add a comment to clarify it.

pkg/controller/platformfeatures_api.go
gainsley: Actually the API is being used by the QA acceptance tests, to see if it should skip the test if the target cloudlet/zone doesn't support k8s or docker.

PR #375, opened 2024-09-24
nbi scaffolding

gainsley:
This adds some scaffolding around how to handle NBI APIs.

We add an echo server to the controller on the existing HTTP API endpoint, shared with the grpc-gateway APIs which aren't actually being used by anything.

We add the `NBIAPIs` object in the controller which will be the handler for all NBI APIs. Currently only the `GET edge-cloud-zones` API is implemented as a test to ensure the basic communication is working. Other APIs will be implemented in later PRs, as implementing everything all at once will be a lot of changes.

We add the `nbictl` library which allows for CLI and e2e test clients to leverage common client code.

We add the `nbiconvert` library which will handle conversion between NBI structs and edgeproto structs.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -548,7 +548,7 @@ func startServices() error {
 	if err != nil {
 		return fmt.Errorf("Failed to create grpc gateway, %v", err)
 	}
-	mux.Handle("/edgeproto/v1/", gw)
+	mux.Handle(cloudcommon.ControllerEdgeprotoRESTPath, gw)
 	// Suppress contant stream of TLS error logs due to LB health check. There is discussion in the community
 	//to get rid of some of these logs, but as of now this a the way around it.   We could miss other logs here but
 	// the excessive error logs are drowning out everthing else.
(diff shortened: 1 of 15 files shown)


pkg/controller/controller.go
```diff
@@ -547,13 +548,21 @@ func startServices() error {
 	if err != nil {
 		return fmt.Errorf("Failed to create grpc gateway, %v", err)
 	}
-	mux.Handle("/", gw)
+	mux.Handle("/edgeproto/v1/", gw)
```
levshvarts, line 551: Did you just want to version our api?

pkg/nbictl/nbictl_data.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+type APIErr struct {
+	Desc   string
+	Status int
+	Err    error
+}
+
+func getURL(addr string) string {
+	return addr + "/edge-application-management/vwip"
```
levshvarts, line 50: Since this is used in several places, might make sense to create a constant for this.

levshvarts (approved):
Thanks for splitting the changes - much easier to understand the logic here.

pkg/controller/controller.go
gainsley: That and it needs to be on a different path than the NBI hander.

PR #376, opened 2024-10-01
azure aks fixes

gainsley:
This fixes our Azure AKS platform. The changes here are:
- AKS platform changes for using service principal instead of user login credentials
- Use AKS SDK (go library) instead of "az" command line client
- AKS platform expects a resource group and uses it for all clusters, instead of creating a resource group for every cluster
- Get credentials returns the kubeconfig instead of writing it to ~/.kube/config (that overwrote my local kubeconfig during testing)
- Managed K8S platforms now honor the crmOnEdge setting
- AKS now only supports crmOnEdge=false
- unit tests and e2e test instructions for testing against AKS

The other two managed kubernetes platforms for GCP/AWS need the same attention, but that can be done in a separate PR. For now they are left somewhat broken.


levshvarts (approved):
Nice! Thanks for the quick fix.

PR #377, opened 2024-10-17
Enhanced resource specifications

gainsley:
This changes our resource specifications away from flavors and towards a more kubernetes-style of individual resources. Previously, we internally used flavors as the resource specification. This did not allow for fine-grained specification of resources for Kubernetes applications, and it did not allow for calculating AppInst resource usage in a ClusterInst (because the AppInst flavor was treated as equivalent to the ClusterInst's flavor).

Additionally, we now support specifying multiple node pools for Kubernetes. There is still more work needed here:
- The infra platforms still only support a single node pool
- The way GPUs and other additional resources are specified needs to be reworked

The old "serverlessconfig", which was a way of specifying kubernetes-like resources, but was converted into a flavor internally, is now deprecated.

This work is done as part of the support for the CAMARA NBI, which specifies resources using individual resources instead of flavors, and requires support for multiple Kubernetes node pools. 

Note that we still support specifying resources via flavors - they are just converted to the new resource specs internally.


pkg/controller/clusterinst_api.go
```diff
@@ -892,61 +968,78 @@ func (s *ClusterInstApi) createClusterInstInternal(cctx *CallContext, in *edgepr
...
 				}
+			} else {
+				log.SpanLog(ctx, log.DebugLevelApi, "using default master node resources")
+				// no master node flavor specified by admin, just
+				// use minimum resources.
+				masterNodeResources = &edgeproto.NodeResources{
+					Vcpus: 1,
+					Ram:   2048,
+					Disk:  10,
+				}
```
levshvarts, line 1034: Do you think this is sufficient to run control plane? For the disk I think we need to have 20G as I don't remember if I optimized the latest base image to fit into 10G

pkg/controller/upgrade_funcs.go
```diff
@@ -81,6 +81,31 @@ func unmarshalUpgradeObj(ctx context.Context, str string, obj interface{}) error
 	return nil
 }
 
+func walkUpgradeObjs[Obj any](ctx context.Context, objStore objstore.KVStore, dbKey string, workFunc func(ctx context.Context, stm concurrency.STM, buf *Obj) error) error {
```
levshvarts, line 84: Thanks - that's useful.

pkg/platform/common/vmlayer/gpu.go
```diff
@@ -330,13 +332,21 @@ var NvidiaGPUOperatorApp = edgeproto.App{
...
 		},
 	},
+	// TODO: what resources does the NVidia operator need?
+	// should those resources be tracked by the controller?
+	KubernetesResources: &edgeproto.KubernetesResources{
+		CpuPool: &edgeproto.NodePoolResources{
+			TotalVcpus:  *edgeproto.NewUdec64(0, 200*edgeproto.DecMillis),
+			TotalMemory: 200,
+		},
+	},
```
levshvarts, line 342: Another tracking issue maybe for this?

levshvarts (approved):
Looks good. Thanks for adding this feature.

pkg/controller/app_api.go
```diff
@@ -1103,15 +1160,24 @@ func (s *AppApi) tryDeployApp(ctx context.Context, stm concurrency.STM, app *edg
...
 func (s *AppApi) ShowZonesForAppDeployment(in *edgeproto.DeploymentZoneRequest, cb edgeproto.AppApi_ShowZonesForAppDeploymentServer) error {
+	// TODO: This function needs a major overhaul. It uses too much
+	// code that is not common with the actual AppInst deploy path.
+	// What it should probably do is take a reference to an existing
+	// App definition, rather than require the user to specify a full
+	// App. That will ensure the App definition is well formed and has
+	// already been validated via App create, and default values filled
+	// in. We should then consider running the actual AppInst create
+	// command with an option to exit before writing any etcd state.
+	// Otherwise this code will just be too hard to maintain.
```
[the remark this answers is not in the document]
gainsley: No, I don't think anyone's using it so it's not worth tracking. If someone wants to use it, then we can open a tracking issue to get it fixed.

pkg/controller/upgrade_funcs.go
gainsley: Yup, trying to reduce boiler plate code.

pkg/platform/common/vmlayer/gpu.go
gainsley: Yeah, let me open a general issue around tracking sidecar/OS resources.

pkg/controller/clusterinst_api.go
gainsley: Filed this issue to track: https://github.com/edgexr/edge-cloud-platform/issues/379

pkg/controller/app_api.go
```diff
@@ -1163,7 +1229,8 @@ func (s *AppApi) ShowZonesForAppDeployment(in *edgeproto.DeploymentZoneRequest,
 			keyOk := false
 			err = s.sync.ApplySTMWait(ctx, func(stm concurrency.STM) error {
 				keyOk = false
-				appInst.VmFlavor = allclds[key]
+				// TODO: fixme
+				//appInst.VmFlavor = allclds[key]
```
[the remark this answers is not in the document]
gainsley: You're right, that comment is pretty lacking. I updated the comment - basically that whole function needs to be rewritten. I noted in the comment what needs to change.

gainsley:
Thanks Lev!

PR #380, opened 2024-10-22
opensource mano cluster integration

gainsley:
This adds support for an [OpenSourceMano](https://osm.etsi.org/) platform instance as a managed kubernetes cluster platform.

I have changed the general platform APIs for Cluster create/update/delete to allow for us to capture infrastructure-specific annotations in our etcd database. For example, OSM assigns a UUID to the cluster after it is created, and this is needed to refer to the cluster for later API calls. In this case, the name we assign to the cluster is unique, so it can be used to look up the ID. But not all platforms may work that way (and having to look up the ID every time is inefficient), and there may be other information about the cluster that we wish to store in etcd that are infra-specific.

Otherwise the integration is pretty straightforward, except for the fact that their OpenAPI spec doesn't actually match the implementation in a couple of places. I believe the code is still in development which is why the spec is out of date.


levshvarts (approved):
Nice! Though not nice that so much of their open-spec is not aligned with the actual code :(
With respect to unit-tests, I guess they run only when you have access to the real osm cluster. Do you think there is a point in mocking it, or since it's not stable there is no point?

gainsley:
Yeah I don't see any point in mocking it since they're actively modifying it.

PR #385, opened 2024-11-12
osmk8s platform flavors from env vars

gainsley:
OSMano does not provide any API to query the flavors supported by the VIM. To allow us to specify per-cloudlet flavors manually, we add an env var to allow specifying the flavors as a JSON string of []edgeproto.FlavorInfo. This is a work-around until OSMano can properly tell us the real flavors.

It turns out updating env vars for an existing CCRM-based Cloudlet did not have any effect, because the env vars are set in the platform object during create, and cached from that point. The changes to resolve that are to flush the CCRM platform object cache if the cloudlet object changes.

Additionally, for changes to a flavor list set by env vars to take effect, we refresh the flavors after a cloudlet update. In general we probably need a better way to keep the flavor list in sync in case the flavors change, but luckily infra flavors do not typically change very frequently.

### DIFF
--- a/pkg/ccrm/ccrm_cloudlet.go
+++ b/pkg/ccrm/ccrm_cloudlet.go
@@ -63,6 +63,13 @@ func (s *CCRMHandler) ApplyCloudlet(in *edgeproto.Cloudlet, stream edgeproto.Clo
 			// CRM will handle it
 			return nil
 		}
+		// Delete platform from cache. This will force re-init of the platform
+		// using the new state of cloudlet (env vars, etc) pulled from etcd.
+		// Note that this also happens via Cloudlet cache update to flush
+		// from all CCRM instances, but we need to (potentially redundantly)
+		// do it here to guarantee it is flushed before CloudletChanged()
+		// function requests for it.
+		s.crmPlatforms.Delete(&in.Key)
 		return s.crmHandler.CloudletChanged(ctx, &in.Key, in, responseSender)
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/ccrm/ccrm_cloudlet.go
```diff
@@ -63,6 +63,13 @@ func (s *CCRMHandler) ApplyCloudlet(in *edgeproto.Cloudlet, stream edgeproto.Clo
 			// CRM will handle it
 			return nil
 		}
+		// Delete platform from cache. This will force re-init of the platform
+		// using the new state of cloudlet (env vars, etc) pulled from etcd.
+		// Note that this also happens via Cloudlet cache update to flush
+		// from all CCRM instances, but we need to (potentially redundantly)
+		// do it here to guarantee it is flushed before CloudletChanged()
+		// function requests for it.
+		s.crmPlatforms.Delete(&in.Key)
```
levshvarts, line 72: Does restarting ccrm pod achieve the same result?

levshvarts (approved):
Looks good. With respect to polling the cloudlet state - we already do that to get metrics, maybe we can add the flavor data there as well.

pkg/ccrm/ccrm_cloudlet.go
gainsley: Yup, it does.

gainsley:
Thanks for reviewing, Lev. Yup we do have some existing periodic frameworks for ccrm, specifically for periodically refreshing the LB certs.

PR #386, opened 2024-11-12
show redacted cloudlet access vars

gainsley:
Rather than completely hiding accessvars set during Cloudlet create/update, this retains the keys so that users can see which access vars they've set.


levshvarts (approved):
Looks good.

PR #388, opened 2024-11-21
show cloudlet and cluster resource apis

gainsley:
This adds two new APIs, ShowCloudletResourceUsage and ShowClusterResourceUsage that show the used and total resources in Cloudlets and Clusters. This allows the user to see exactly what resources the system believes are in use. This is important both for users, and also us as developers to ensure that the algorithms that are choosing cloudlets/clusters based on available resources are working as expected.

We already had an API to get a single cloudlet's resource usage, so that is just extended to a new show command for all cloudlets. The resourceScore functions were just moved to different files to be easier to use.


pkg/controller/cloudlet_resource_check.go
```diff
@@ -381,3 +382,43 @@ func (s *CloudletResCalc) getResourceLimits() ResLimitMap {
...
+// on a cloudlet. A higher score means more available resources.
+func (s *CloudletResCalc) calcResourceScore(usedVals resspec.ResValMap) uint64 {
+	// get max value for each resource on cloudlet
+	maxVals := getMaxResourceVals(s.deps.cloudletInfo.ResourcesSnapshot.Info, s.deps.cloudlet.ResourceQuotas)
+	// Calculate score based on weights and free values
+	// Because some resources may have no limit, track the number
+	// of resources we've scored. We'll divide by this number to
+	// get an average per-resource score for comparisons.
+	var score, numScored uint64
+	for res, weight := range resourceWeights {
```
levshvarts, line 396: `resourceWeights` is in `potential_cloudlet.go`, but not even used there anymore. I suggest moving it to `cloudcommon` pkg, since map values are from that package anyways.

pkg/controller/cluster_resources.go
```diff
@@ -169,3 +170,160 @@ func NodePoolFromResources(name string, pr *edgeproto.NodePoolResources) *edgepr
 		NodeResources: &nr,
 	}
 }
+
+// calcResourceScore gets a score which represents the available resources
+// in a cluster. A higher score means more available resources.
+func (s *ClusterInstApi) calcResourceScore(free resspec.ResValMap) uint64 {
```
levshvarts, line 176: Eh, logic is so close between calculation from free and used...Maybe call them `calcResourceScoreFromFree` and `calcResourceScoreFromUsed` to mark the difference?

levshvarts (approved):
Looks good.

tools/protogen/protogen.proto
```diff
@@ -200,6 +200,9 @@ extend google.protobuf.MethodOptions {
   string cli_cmd = 51050;
   // Custom input validator for mc
   bool mc2_custom_validate_input = 51061;
+  // Generate a hook function for customizing the authorization behavior of the
+  // API in MC for custom show APIs.
+  bool mc2_show_custom_authz = 51067;
```
levshvarts, line 205: I don't see this method option used in any of the new methods though...

pkg/controller/cloudlet_resource_check.go
gainsley: Hm, I can move it to cloudlet_resource_check.go. I want to keep all the decision-making algorithm code inside the controller (or some specific decision alg package), as none of the other services should be making decisions about deployment placement.

levshvarts: Ok, I see. In that case it doesn't really matter where that variable is declared - can just stay in potential_cloudlet.go

tools/protogen/protogen.proto
gainsley: Whoa, good catch! It's actually supposed to be in the ShowClusterResourceUsage method.

levshvarts (approved):
Looks good. Thanks for fixing this.

pkg/controller/cluster_resources.go
levshvarts: I am talking about logic in `(s *CloudletResCalc) calcResourceScore()`

gainsley: Oh I see, one takes freeRes args and the other used. I'll rename the CloudletResCalc one to calcResourceScoreFromUsed, as in general resource scores are based on free resources.

PR #389, opened 2024-11-21
restore k8s appinst manifest on delete

gainsley:
When using the CCRM, a restart/upgrade of the CCRM service will cause any locally stored k8s manifests to be lost. For CRMs these remain after restart.

Because these are needed for delete, on delete we re-write the manifest to ensure that it is present in case it was lost.

### DIFF
--- a/pkg/k8smgmt/appinst.go
+++ b/pkg/k8smgmt/appinst.go
@@ -449,7 +449,10 @@ func UpdateAppInst(ctx context.Context, accessApi platform.AccessApi, client ssh
 	return WaitForAppInst(ctx, client, names, app, WaitRunning)
 }
 
-func DeleteAppInst(ctx context.Context, client ssh.Client, names *KubeNames, app *edgeproto.App, appInst *edgeproto.AppInst) error {
+func DeleteAppInst(ctx context.Context, accessApi platform.AccessApi, client ssh.Client, names *KubeNames, app *edgeproto.App, appInst *edgeproto.AppInst) error {
+	if err := WriteDeploymentManifestToFile(ctx, accessApi, client, names, app, appInst); err != nil {
+		return err
+	}
 	configDir := getConfigDirName(names)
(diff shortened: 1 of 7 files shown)


levshvarts (commented):
This change looks good. However when we create appInst, we apply the whole directory and purge anything that doesn't exist there in `createOrUpdateAppInst`. We might need to handle that as well, as for ccrm restart the directory won't exist.

gainsley:
> This change looks good. However when we create appInst, we apply the whole directory and purge anything that doesn't exist there in `createOrUpdateAppInst`. We might need to handle that as well, as for ccrm restart the directory won't exist.

Thanks Lev for reviewing. If I understand this correctly, your worry is that create AppInst may purge stuff from the cluster that it shouldn't, like other Apps in the same cluster? I agree that's a valid concern. I think that can also happen if there are 2 CCRMs, and two Apps are added to the same cluster, but those createAppInst calls land on different CCRM instances. We need to change our purge selector to be per AppInst, which it isn't right now. I'll file a bug for that.

levshvarts:
> > This change looks good. However when we create appInst, we apply the whole directory and purge anything that doesn't exist there in `createOrUpdateAppInst`. We might need to handle that as well, as for ccrm restart the directory won't exist.
> 
> Thanks Lev for reviewing. If I understand this correctly, your worry is that create AppInst may purge stuff from the cluster that it shouldn't, like other Apps in the same cluster? I agree that's a valid concern. I think that can also happen if there are 2 CCRMs, and two Apps are added to the same cluster, but those createAppInst calls land on different CCRM instances. We need to change our purge selector to be per AppInst, which it isn't right now. I'll file a bug for that.

Yeah - I've hit that before several times. Can you file a tracking issue on that if we don't have one?

gainsley:
> Yeah - I've hit that before several times. Can you file a tracking issue on that if we don't have one?

https://github.com/edgexr/edge-cloud-platform/issues/391

PR #394, opened 2024-12-03
add k8ssite platform: k8s-cluster-as-a-cloudlet

gainsley:
This modifies the old k8sop platform to a new k8ssite platform - the intent is the same, that a single kubernetes clusters acts as the entire cloudlet. The difference is the old k8sop platform assumed that CRM would be deployed in the cluster, while the new k8ssite platform uses the CCRM running in the regional control-plane cluster. So the old k8sop files have become the new k8ssite files (although the diff doesn't really reflect that).

The AppInst-handling code between k8sop-appinst and mk8s-appinst was the same, so I've refactored that out into the k8swm (workload manager) package.

Other than that there are a few fixes here and there.


pkg/controller/cloudlet_api.go
```diff
@@ -663,6 +663,9 @@ func (s *CloudletApi) createCloudletInternal(cctx *CallContext, in *edgeproto.Cl
 		in.RootLbFqdn = getCloudletRootLBFQDN(in)
 		in.StaticRootLbFqdn = in.RootLbFqdn
 		if features.IsSingleKubernetesCluster {
+			if in.EnableDefaultServerlessCluster {
+				return errors.New("cannot enable default serverless cluster for single kubernetes cluster platform")
+			}
```
levshvarts, line 668: I am confused - aren't these two equivalent? Or a single k8s clusters can be defaultServerlessCluster

pkg/k8smgmt/manifest.go
```diff
@@ -339,7 +340,15 @@ func MergeEnvVars(ctx context.Context, accessApi platform.AccessApi, app *edgepr
...
 			return kubeManifest, fmt.Errorf("unable to marshal the k8s objects back together, %s", err.Error())
 		} else {
-			files = append(files, buf.String())
+			file := buf.String()
+			if _, ok := o.(*networkingv1.NetworkPolicy); ok {
+				// NetworkPolicyStatus has been removed as of
+				// https://github.com/kubernetes/api/commit/90ceadb2d5f2f1d135492b647c9fb72777db4b36
+				// unfortunately yaml printer writes it as an empty {}
+				// field, we need to remove it
+				file = strings.TrimSuffix(file, "status: {}\n")
```
levshvarts, line 349: good thing you tested MT cluster now :)

pkg/platform/platform_names.go
```diff
@@ -37,7 +37,7 @@ const (
 	PlatformTypeFederation        = "federation"
 	PlatformTypeGCP               = "gcp"
 	PlatformTypeK8SBareMetal      = "k8sbaremetal"
-	PlatformTypeK8SOperator       = "k8soperator"
+	PlatformTypeK8SSite           = "k8ssite"
```
levshvarts, line 40: better name!

levshvarts (approved):
Looks good. Would've been better if github could notice the difference between k8sop-* and k8ssite-* files rather than just "delete one file, add completely new one".

pkg/controller/cloudlet_api.go
gainsley: They're somewhat different. The defaultCluster is a "fake" cluster that represents the entire cloudlet. Because the cloudlet platform doesn't actually support CreateCluster, we create this cluster via direct STMPut into Etcd. The defaultMTCluster is a MT cluster meant to co-exist with other dedicated-tenant clusters on a Cloudlet that supports CreateCluster, so it gets built the normal way. Resource handling is also slightly different, for the solo case, the platform dedicates all resources to the fake cluster. For the MT case, it just gets some default set of resources.

pkg/k8smgmt/manifest.go
gainsley: Yeah, good to work out these issues now :)

pkg/controller/cloudlet_api.go
levshvarts: Thanks for the explanation.

PR #398, opened 2024-12-11
fix reservable autocluster cleanup on undo

gainsley:
Testing found an issue with the cleanup of a reservable autocluster created during CreateAppInst, if the AppInst create failed. The problem was that the AppInst API finished before the cluster was fully deleted. This caused subsequent tests to fail.

The problem stemmed from the streamObj handling. When the AppInst create fails, it sets the undo flag on the context and calls deleteClusterInstInternal. The undo flag triggers streamStart() to try to reuse the existing stream, but this process did not carry over the crmMsgCh. Without the crmMsgCh, the call to WaitForCloudletInfo in deleteClusterInstInternal returns early, instead of waiting until the clusterinst is deleted.

Using the Undo flag is tricky, because in this case, the stream is technically not resuable by deleteClusterInstInternal, because we are actually undoing the AppInst create. The stream used by createClusterInstInternal to create the autocluster was already finished, and the only stream present is the AppInst stream. So, we now track streams by their streamKey on the CallContext. This allows stream reuse to trigger off the presence of the stream to reuse, rather than the undo flag.

Because crmMsgCh is not really optional to the WaitFor funcs (there's no point in calling it if the channel is nil), I've made it a required parameter instead of optional.

### DIFF
--- a/api/edgeproto/appinst.pb.go
+++ b/api/edgeproto/appinst.pb.go
@@ -13,6 +13,7 @@ import (
 	"github.com/edgexr/edge-cloud-platform/pkg/objstore"
 	"github.com/edgexr/edge-cloud-platform/pkg/util"
 	_ "github.com/edgexr/edge-cloud-platform/tools/protogen"
+	"github.com/go-redis/redis/v8"
 	_ "github.com/gogo/googleapis/google/api"
 	_ "github.com/gogo/protobuf/gogoproto"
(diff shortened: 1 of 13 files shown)


levshvarts (approved):
Looks good. Must have been a difficult one to debug....

gainsley:
Yeah it was a bit annoying to debug...but a slew of debug prints always works :)

PR #402, opened 2024-12-17
infra prop for ingress controller present

gainsley:
Add common infra property to control whether or not to install ingress-nginx. Some clusters (rke2) may be given to us with the ingress controller already present.

Added a dockerfile for an image with a proper wget installed that handles proxies (the default wget in alpine does not).

I also added checks on Cloudlet.EnvVars to match checks on Cloudlet.AccessVars. This validates the env vars are spelled properly and actually supported.

### DIFF
--- a/build/docker/Dockerfile.alpine-wget
+++ b/build/docker/Dockerfile.alpine-wget
@@ -1,0 +1,5 @@
+
+FROM alpine
+
+USER root
+RUN apk add --no-cache wget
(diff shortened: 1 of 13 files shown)


levshvarts (approved):
Changes look good. One question - with a provided ingress, won't we need a different way of apply configurations to it?

gainsley:
In general all ingress controllers should work with a standard Kubernetes Ingress spec, assuming we're not using any custom labels to change the behavior. We do use `nginx.ingress.kubernetes.io/ssl-redirect`, but that just forces a redirect from HTTP to HTTPS, otherwise client just gets a failed connection for HTTP, so it's not necessary.

PR #403, opened 2024-12-17
mark cluster resources if reserved by org

gainsley:
Mark clusterResourceUsage with reservedBy for reservable clusters. This allows RBAC to determine if a user can see resources for a cluster they reserved.


levshvarts (approved):
Looks good.

PR #404, opened 2024-12-17
fix clusterinst reservation handling for multiple appinsts

gainsley:
We recently made changes to allow AppInst deployments to target an existing reservable ClusterInst if already owned by the same organization, and scale that clusterinst up to fit the second AppInst. This fixes a bug where deleting the second AppInst would clear the reservation, even though the first AppInst was still present.


levshvarts (approved):
That's a good catch!

PR #408, opened 2024-12-18
potential cloudlet skip if not enough resources

gainsley:
For AppInst creates, if no existing cluster was found, we picked the first potential cloudlet from the list, sorted by resource score (i.e. with the most resources available). However, this didn't actually check if the specific resources needed were available. A cloudlet can have a high resource score if other resources are abundant, but if there aren't enough vcpus, the cluster can't be created.

We now check that cloudlets can fit the desired auto cluster instance.


levshvarts (approved):
Looks good.

PR #409, opened 2024-12-18
various k8s appinst deploy fixes

gainsley:
Various fixes:
- AppInst compatibility "RegionScopedNames" which was implemented a long time ago was actually never used, because all new AppInsts were still using the previous compatibility version. This updates new AppInsts to use the new compatibility version.
- For AppInsts with ingress, we need to make sure the URI is based on the AppInst name, not a shared LB name. Otherwise, when deploying multiple AppInsts to the same cluster, the hostnames in the ingress objects for each AppInst will conflict because they are the same.
- For ClusterResourceUsage Show, if one cluster hits an error, skip it instead of failing the entire API
- k8smgmt.CheckPodsStatus is reworked. Previously, it was based on kubectl output that used the "Reason" field as the status, but the reason field is an arbitrary string, and is not reliable to expect certain strings. Because of this it sometimes failed with new types of managed clusters. It now checks fields and struct data that are not arbitrary.
- k8smgmt.createOrUpdateAppInst now has baked in undo and wait (optional). Some callers were implementing their own undo, and some had incorrectly omitted it. Most callers were calling wait inline, which is redundant. Only callers that are doing wait in parallel now specify the NoWait() option
- ingress object manifest files are now written to the cluster's manifest config dir, otherwise they end up getting pruned and deleted if another change is applied (AppInst update, or another AppInst is created).

### DIFF
--- a/pkg/controller/appinst_api_test.go
+++ b/pkg/controller/appinst_api_test.go
@@ -922,19 +922,19 @@ func testSingleKubernetesCloudlet(t *testing.T, ctx context.Context, apis *AllAp
 	}, {
 		"MT specified correct cluster",
 		0, &zoneMT, mtOrg, mtClust, notDedicatedIp,
-		"shared.singlek8smt-unittest.local." + appDnsRoot, PASS,
+		"pillimogo1-atlanticinc.singlek8smt-unittest.local." + appDnsRoot, PASS,
 	}, {
 		"MT any clust name blank org",
 		0, &zoneMT, "", "", notDedicatedIp,
-		"shared.singlek8smt-unittest.local." + appDnsRoot, PASS,
+		"pillimogo1-atlanticinc.singlek8smt-unittest.local." + appDnsRoot, PASS,
 	}, {
 		"ST specified correct cluster",
 		0, &zoneST, stOrg, stClust, notDedicatedIp,
-		"shared.singlek8sst-unittest.local." + appDnsRoot, PASS,
+		"pillimogo1-atlanticinc.singlek8sst-unittest.local." + appDnsRoot, PASS,
 	}, {
 		"ST blank clust name blank org",
 		0, &zoneST, "", "", notDedicatedIp,
-		"shared.singlek8sst-unittest.local." + appDnsRoot, PASS,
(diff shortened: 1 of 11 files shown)


pkg/controller/appinst_api_test.go
```diff
@@ -922,19 +922,19 @@ func testSingleKubernetesCloudlet(t *testing.T, ctx context.Context, apis *AllAp
 	}, {
 		"MT specified correct cluster",
 		0, &zoneMT, mtOrg, mtClust, notDedicatedIp,
-		"shared.singlek8smt-unittest.local." + appDnsRoot, PASS,
+		"pillimogo1-atlanticinc.singlek8smt-unittest.local." + appDnsRoot, PASS,
```
levshvarts, line 925: These changed test cases are all HTTP-based apps? It's hard to cross-reference with which appinst this is. I looked at `AppData()[0]`, but this doesn't look like it should get a dedicated fqdn....

pkg/k8smgmt/appinst.go
```diff
@@ -60,86 +95,54 @@ func LbServicePortToString(p *v1.ServicePort) string {
...
-					// pod's container is in state Terminated,
-					// before it's actually removed. If we hit this
-					// while waiting for a delete, let it try again
-					continue
-				}
-				if strings.Contains(podState, "Init") {
-					// Init state cannot be matched exactly, e.g. Init:0/2
-					log.SpanLog(ctx, log.DebugLevelInfra, "pod in init state", "podName", podName, "state", podState)
+	podList := v1.PodList{}
+	if err := json.Unmarshal([]byte(out), &podList); err != nil {
```
levshvarts, line 105: nice - finally structured data not string-parsing :)

levshvarts (approved):
All really good fixes! Thanks a lot!

pkg/controller/appinst_api_test.go
gainsley: Yeah, they all use AppData()[0] which has an http accessport.

pkg/k8smgmt/appinst.go
gainsley: I know, this thing was super flakey. This code is btw based off of the same function in our edgecloud operator, so it's pretty well tested already.

PR #411, opened 2024-12-20
kubectl version, use cert with existing ingress controller

gainsley:
A few more fixes:
- Update the kubectl wrapper in the build, this new version allows for using https proxies
- for Federation of a single cluster-as-a-cloudlet zone, owned by a tenant, we want some way to allow federated tenants to deploy. We need a better process here, but for now we check if the remote tenant is the same as the local tenant owner of the cloudlet.
- For clusters that already have an ingress controller installed, we should be able to use that ingress controller with our own certs. It's a little complicated because we only want to install one cert shared by all ingresses, but Kubernetes requires that the cert is in the same namespace as the ingress. For now we only support the cert in the default namespace. Future works needs to be done to maintain copies of the cert in all namespaces that are created/used by AppInsts.


levshvarts (approved):
Thanks for the fixes.

gainsley:
Thanks for reviewing

PR #412, opened 2024-12-21
more informational messages for appinst deploy algorithms

gainsley:
This provides more transparency as to why some cloudlets or clusters may not be able to be used when looking for a way to deploy an AppInst. We provide messages regarding the potential cloudlets and reasons that some may be skipped. Messages are intended to be generic rather than specific, as we are not supposed to expose the cloudlet-specific information to the user.

An example:
```
message: Found 3 potential cloudlets for deployment
message: Found 0 potential clusters for deployment, some skipped because instance of App already present
message: No existing cluster available
message: Creating new auto-cluster named reservable0-1aa80307 to deploy AppInst
```
Failure case (no existing available and no resources to create a new one):
```
message: Found 1 potential cloudlets for deployment
message: Found 0 potential clusters for deployment, some skipped because k8s version not supported
message: No existing cluster available
Error: No resources available for deployment
```

Also an unrelated change: for k8ssite platforms, figure out the version of the underlying kubernetes cluster.

### DIFF
--- a/pkg/cloudcommon/names.go
+++ b/pkg/cloudcommon/names.go
@@ -266,6 +266,7 @@ const (
 	AnnotationBadUpgrade55Name        = "bad-upgrade55-name"
 	AnnotationPreviousDNSName         = "previous-dns-name"
 	AnnotationFedPartnerAppProviderID = "fed-partner-app-provider-id"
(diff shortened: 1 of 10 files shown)


levshvarts (approved):
Changes look good.

PR #413, opened 2024-12-21
add cloudlet refreshcerts api

gainsley:
Add API to trigger refreshing cloudlet certs. This only applies to CCRM-based cloudlets.


pkg/controller/cloudlet_api.go
```diff
@@ -2857,3 +2857,66 @@ func redactAccessVars(accessVars map[string]string) map[string]string {
 	}
 	return vars
 }
+
+func canRunRefreshCerts(cloudlet *edgeproto.Cloudlet, features *edgeproto.PlatformFeatures) bool {
```
levshvarts, line 2861: Do you need `feature`  argument, or you just wanted to make this generic?

pkg/controller/cloudlet_api.go
```diff
@@ -2857,3 +2857,66 @@ func redactAccessVars(accessVars map[string]string) map[string]string {
...
+
+func (s *CloudletApi) RefreshCerts(key *edgeproto.CloudletKey, cb edgeproto.CloudletApi_RefreshCertsServer) error {
+	ctx := cb.Context()
+	ptof := s.all.platformFeaturesApi.FeaturesByPlatform()
+
+	filter := &edgeproto.Cloudlet{
+		Key: *key,
+	}
+	cloudlets := []edgeproto.Cloudlet{}
+	err := s.cache.Show(filter, func(obj *edgeproto.Cloudlet) error {
```
levshvarts, line 2874: Can this be a partial key? If this is a complete cloudlet key, there is only a single cloudlet to consider and `s.Get()` would be a better fit.

levshvarts (approved):
Looks good.

pkg/controller/cloudlet_api.go
gainsley: I just wanted to make it generic, as some platforms may not need to cert refresh.

pkg/controller/cloudlet_api.go
gainsley: Yup, it's intended to allow for a partial key specified by the user.

PR #415, opened 2025-01-13
Osmk8s: cluster get-creds fixes

gainsley:
Fix how OSM K8s cluster credentials are retrieved. Apparently, the client must run two API calls to get credentials.

Also a fix for updating cloudlet flavors, where the updated flavors were not being updated to the cloudlet object if the new flavors were the same number as the old flavors.


levshvarts (approved):
Looks good.

PR #416, opened 2025-01-22
osm code reorg, add osm workload manager

gainsley:
This reorganizes most of the OSM code into an `osmclient` package. The osclient is then used by the osmk8s package to manage clusters, and the osmwm (workload manager) to manage apps.

Because of the fact that the app workload manager really only deploys apps, and doesn't handle setting up ingress, certs, and DNS, I have split what used to be the k8s workload manger into a parent k8sPlatformMgr, which handles ingress/certs/DNS, and an inner "WorkloadMgr" interface which is implemented either by our k8smgmt code, or the OSM client. The inner workload manager deploys applications, but does not deal with the ingress/certs/DNS.

There is new code here for the osmclient to handle creating apps and app instances.


pkg/platform/osmano/osmclient/osm-app.go
```diff
@@ -0,0 +1,287 @@
...
+
+var appT = template.Must(template.New("appT").Parse(appTemplate))
+
+var appTemplate = `apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ .Namespace }}
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1
+kind: Kustomization
```
levshvarts, line 274: Is this basically a flux app definition, or some CRD they wrote themselves?

levshvarts (approved):
Thanks for refactoring - looks much cleaner now!

pkg/platform/osmano/osmclient/osm-app.go
gainsley: This is a FluxCD CRD, but the parameterization is something specific to OSM. I don't understand all the details, but it looks like the OSM APIs trigger ArgoCD workflows to run their shell scripts to process the data, commit that manifests to their internal git repo, and then push a FluxCD kustomization CRD to the target cluster. FluxCD then runs in the target cluster, pulling from the git repo, and applying the changes (actually deploying the AppInst).

gainsley:
Hey @levshvarts , I have updated this PR with two more commits (the last commit is just removing some junk). This cleans up some of the duplicate code between the k8sWorkloadMgr and the OSMWorkloadMgr and reorgs the code a bit more, now that I have a better understanding of what the OSM workload manager can do.

I think it's still a bit overly complex, but the division of labor is:

k8spm.K8sPlatformMgr (and vmlayer.VMProvider)
- handles writing kubeconfigs, docker registry secret, DNS and Ingress
- calls into k8smgmt code: k8smgmt.CreateAppInst()
k8smgmt code
- handles namespaces, network policy, resource quota, takes as input an optional k8smgmt.WorkloadMgr
- calls the workloadMgr ApplyAppInstWorkload()
k8smgmt.K8sWorkloadMgr or osmwm.OSMWorkloadMgr
- deploys the AppInst "workload" (just deploys what is defined for the App).

The division between the WorkloadMgrs and the k8smgmt code is necessitated due to wanting to swap between different "Workload Managers", even though what they're doing is pretty basic.

The division between K8sPlatformMgr/VMProvider and k8smgmt was I think to separate platform-specific code from the platform-independent k8smgmt code, but I'm not really sure how much of that is really platform specific. Besides creating the ssh.Client that points to the container/VM on which to write the files, and the patching of the VM's node IP as the DNS IP in VMProvider, I think the rest could be probably be moved to k8smgmt. But needs some investigation, and not really necessary at this point.

Please take a look when you get a chance. Thanks.

pkg/k8spm/k8spm.go
```diff
@@ -80,13 +86,17 @@ func (m *K8sWorkloadMgr) CreateAppInst(ctx context.Context, clusterInst *edgepro
 	}
 
 	updateSender.SendStatus(edgeproto.UpdateTask, "Deploying App Instance")
-	switch deployment := app.Deployment; deployment {
+	// TODO: We should handle helm the same as a
+	// manifest-based appinst inside the workload manager,
+	// so that the surrounding handlers for namespaces and
+	// policies are common to both.
```
levshvarts, line 92: Can you open an issue for this. I think this is a good idea, but we'll forget unless we track this.

levshvarts (approved):
Looks good.

pkg/k8spm/k8spm.go
gainsley: Yep, I had opened https://github.com/edgexr/edge-cloud-platform/issues/420

PR #419, opened 2025-01-27
deploy k8s appinsts in their own namespace

gainsley:
This change deploys Kubernetes AppInsts to their own namespace. Previously, only AppInsts in multi-tenant clusters were deployed to their own namespaces. Deploying AppInsts to their own namespace makes it easier to manage multiple objects created by the AppInst, and makes it easier to assign resource limits by limiting the resources consumed by the namespace. In this vein the resource limits are now applied to the namespace rather than to the deployments in the manifest. This is a better more general solution that can also allow us to limit resources for helm Apps.


levshvarts (approved):
Looks good.

PR #422, opened 2025-01-28
add back pod resource limits, only apply ResourceQuota for multi-tenancy

gainsley:
This fixes an issue with change https://github.com/edgexr/edge-cloud-platform/pull/419.

If a podTemplateSpec does not have any resource limits defined, then the presence of a namespace ResourceQuota will prevent that pod from being deployed:
```
Error: Create App Inst failed: no pods found, replicaset statuses: app110-deployment-7f7c58d898: pods "app110-deployment-7f7c58d898-s75fh" is forbidden: failed quota: app1-acmeappco: must specify limits.cpu for: app110; limits.memory for: app110
```

To avoid this, we
- add back the resource limits on the pod template spec
- only apply ResourceQuota in a multi-tenant context
- check the replicaset status if no pods are found (that's where the error is)


levshvarts (approved):
Looks good.

PR #424, opened 2025-01-30
fix missing commercial certs setting

gainsley:
When creating a new CCRM-based cloudlet, we pass a bunch of settings to the platform-specific CreateCloudlet via the edgeproto.PlatformConfig, which gets converted to platform.PlatformConfig. The CommercialCerts setting was not being copied over, resulting in a new Cloudlet using self-signed ingress certs instead of proper certs.

This properly fixes the copying of the setting.


levshvarts (approved):
Looks good.

PR #427, opened 2025-02-05
Vendor-specific tags for apps and clusters.

levshvarts:
In order to support special apps, we need to add tags field.
In addition, I added code to facilitate inference engine service for e2e.


pkg/process/process_vault.go
```diff
@@ -70,20 +70,22 @@ type VaultRoles struct {
...
+	ClusterSvcRoleID     string `json:"clustersvcroleid"`
+	ClusterSvcSecretID   string `json:"clustersvcsecretid"`
+	EdgeTurnRoleID       string `json:"edgeturnroleid"`
+	EdgeTurnSecretID     string `json:"edgeturnsecretid"`
+	AutoProvRoleID       string `json:"autoprovroleid"`
+	AutoProvSecretID     string `json:"autoprovsecretid"`
+	FrmRoleID            string `json:"frmroleid"`
+	FrmSecretID          string `json:"frmsecretid"`
+	InferenceSvcRoleID   string `json:"inferencesvcroleid"`
+	InferenceSvcSecretID string `json:"inferencesvcsecretid"`
```
gainsley, line 88: I think the inference service can just use the cluster-svc role/secret IDs. They share the same effective role. Btw I assume the inference-svc will run in the regional cluster?

levshvarts: correct, will run at the region level, since the apps are per-region.

gainsley (approved):

PR #429, opened 2025-02-10
InferenceModel show api

levshvarts:
Added `/show/inferencemodel` api for prebuilt model apps.


gainsley (commented):
Hey Lev, do we really need a separate API for this? What about if someone adds another custom app provider, or another type of apps besides Inference? Do we need to another API for that? It seems like this approach doesn't scale well if we need to add APIs for every single app type.
Is the issue that the apps don't show up under the usual "app show" because they are from a different organization? If that's the case I think we need a way to mark Apps as "public" so that any organization can see and use those apps. This would be a better general solution.

levshvarts:
> Hey Lev, do we really need a separate API for this? What about if someone adds another custom app provider, or another type of apps besides Inference? Do we need to another API for that? It seems like this approach doesn't scale well if we need to add APIs for every single app type. Is the issue that the apps don't show up under the usual "app show" because they are from a different organization? If that's the case I think we need a way to mark Apps as "public" so that any organization can see and use those apps. This would be a better general solution.

I was thinking about that. I don't know if it makes sense to have `/show/app` show pre-built public  apps by default. I could add public orgs to the list of all available apps, so the pre-built models are available to all, and this way we could filter based on the tags set, but that still doesn't solve the problem of showing all the prebuilt apps by default.

gainsley:
Good point, I think it makes sense to avoid cluttering a user's private apps with a potentially large collection of public apps. So then I think it makes sense to have a single API for showing public apps, but I would make it generic, i.e. publicapps show, as opposed to a specific "inference" app type.

gainsley:
In the UI then we would have a different section for "Apps" vs "Public Apps"

gainsley (approved):

PR #430, opened 2025-02-10
osm platform final fixes, mock managed k8s platform

gainsley:
Changes:
- OSM KSU requires the AppName variable in the kustomization template.
- OSM KSU can be deployed to a namespace using the "targetNamespace" key in the kustomization template, rather than having add the namespace to every object in the manifest.
- OSM Cluster delete API changed, it now returns 204 instead of 202.
- Fixes to allow scaling cluster via mcctl cluster update command. The OSM code was rejecting it because of extra update flags that can be ignored
- Use a context variable to track undo for k8smgmt functions. This avoids an issue where if the create failed, in the undo path, the delete will also fail when waiting for the pod status, which caused the remaining parts of the delete (clean up the ingress and namespace) to not be run, leaving extra junk behind
- Added a mock managed k8s platform, which is the managed k8s equivalent to the existing mock platform (a VM-based platform)
- Fixed aliases for zone refs for mcctl


levshvarts (approved):
Looks good.

PR #434, opened 2025-03-12
unit tests: fix timeouts, wsl ip

gainsley:
Changing my test setup on Ubuntu WSL2 from NAT networking mode to mirrored networking mode seems to have changed the way some client connections behave when connecting to non-existent services. Previously I assume these used to fail immediately, but now they block indefinitely.

To fix, we configure client timeouts for these connections, and reduce the timeouts for unit-tests since the services are not present.

Also in mirrored networking mode, the local IP may not be on eth0 (mine is on eth4), as windows maps in multiple interfaces to WSL.


pkg/process/process_dockergeneric.go
```diff
@@ -105,12 +105,29 @@ func AddHostDockerInternal(args []string) ([]string, error) {
...
+		if err == nil && strings.TrimSpace(string(out)) == "mirrored" {
+			args = append(args, "--add-host", "host.docker.internal:host-gateway")
+			return args, nil
+		}
 		// get wsl ip
-		out, err = exec.Command("sh", "-c", `ip addr show eth0 | grep -oP '(?<=inet\s)\d+(\.\d+){3}'`).CombinedOutput()
-		if err != nil {
-			return args, fmt.Errorf("Unable to determine WSL ip address, %s, %v", string(out), err)
+		ip := ""
+		for ii := range 10 {
```
levshvarts, line 117: ew...wsl makes this a bit gnarly :)

levshvarts (approved):
Looks good. Thanks for fixing these.

PR #437, opened 2025-04-01
rework gpu resources

gainsley:
This reworks the way we handle gpu resources. Previously gpus were defined via obscure strings like "gpu:t4:1" or "pci_passthrough:t4" based on tags on Openstack flavors. This new change instead uses gpu objects with a model_id, vendor, and count. These objects are pulled from kubernetes node tags, or from infrastructure-specific flavors. In general, it is up to the infrastructure platform code to transform the underlying infra-specific specification of gpus into the general gpu objects.

GPUs can now be queried by zone, and shown as separate cloudlet resources. The zone GPU APIs are meant to be used be developers and hide the actual underlying infra resource counts. The cloudlet GPU APIs are meant to be used by operators to track GPU usage across their infra.

We continue to still honor GPUs specified using restagtables and optresmaps, but in general that approach is deprecated.


pkg/controller/zone_api.go
```diff
@@ -178,3 +181,92 @@ func (s *ZoneApi) getZoneByID(ctx context.Context, id string) (*edgeproto.Zone,
...
+			} else {
+				max = int(limit.InfraMaxValue)
+			}
+		}
+		used := 0
+		if res, ok := usedVals[resKey]; ok {
+			used = int(res.Value.Whole)
+		}
+		// set gpu count to 1 if there are gpus available
+		if max == 0 || max > used {
```
levshvarts, line 257: If max is 0, do we just assume that number of gpus is unlimited?

pkg/k8smgmt/cluster.go
```diff
@@ -317,6 +319,9 @@ func GetNodeInfos(ctx context.Context, client ssh.Client, kconfArg string) ([]*e
 			nodeInfo.Allocatable[name] = dec
 		}
 		for res, quantity := range item.Status.Capacity {
+			if res == cloudcommon.KubernetesAMDGPUResource {
+				amdGPUCount = int(quantity.Value())
+			}
```
levshvarts, line 324: Even though this is different from nvidia, might worth it to put it into getNodeGPUInfo, and let getNodeGPUInfo take `item`. That way at least the count calculation is in the same function.

pkg/k8smgmt/cluster.go
```diff
@@ -394,6 +413,72 @@ func convertNodeResource(res v1.ResourceName, quantity resource.Quantity) (strin
 	return name, dec, nil
 }
 
+func getNodeGPUInfo(labels map[string]string) ([]*edgeproto.GPUResource, *edgeproto.GPUSoftwareInfo, error) {
+	// Note: we only support one GPU per node, as not sure
+	// what the labels would look like for multiple GPUs
+	// or if the gpu-operator even supports it.
```
levshvarts, line 419: `nvidia.com/gpu.count` would be > 1 for that node output. But this should work transparently. I think it's safe to assume multiple gpus that are the same.

pkg/k8smgmt/cluster.go
```diff
@@ -394,6 +413,72 @@ func convertNodeResource(res v1.ResourceName, quantity resource.Quantity) (strin
...
+			gpu.Vendor = cloudcommon.GPUVendorNVIDIA
+			gpuFound = true
+		case "nvidia.com/gpu.memory":
+			val, err := strconv.ParseUint(v, 10, 64)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to parse gpu memory, %s, %v", v, err)
+			}
+			// nvidia memory is specified in MB
+			gpu.Memory = val / 1024
+		case "nvidia.com/gpu.replicas":
```
levshvarts, line 436: `gpu.replicas` might be misleading, as you can change the replicas and that would allow for time-sharing of the gpu. Do we want to support this? The metrics for the gpu and the attribution would be skewed. I definitely think we should track replicas to support time slicing, but if we want to assign whole gpus, `gpu.count` might work better.

pkg/platform/azure/azure.go
```diff
@@ -180,9 +186,63 @@ func (a *AzurePlatform) GatherCloudletInfo(ctx context.Context, info *edgeproto.
 					if err != nil {
 						return fmt.Errorf("failed to parse %s value %s for flavor %s, %s", *cap.Name, *cap.Value, *v.Name, err)
 					}
-					flavor.PropMap = map[string]string{
-						"gpu": fmt.Sprintf("gpu:%d", num),
+					// See https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/overview?tabs=breakdownseries%2Cgeneralsizelist%2Ccomputesizelist%2Cmemorysizelist%2Cstoragesizelist%2Cgpusizelist%2Cfpgasizelist%2Chpcsizelist#gpu-accelerated
```
levshvarts, line 189: I guess this is current naming....and we'll need to update this periodically.

pkg/resspec/vmspec_matcher.go
```diff
@@ -420,9 +416,37 @@ func GetVMSpec(ctx context.Context, nodeResources *edgeproto.NodeResources, cli
 	if skippedExtraRes > 0 {
 		reasons = append(reasons, fmt.Sprintf("%d with optional resources not requested", skippedExtraRes))
 	}
+	for resName, count := range skippedExtra {
+		reasons = append(reasons, fmt.Sprintf("%d with %s not requested", count, resName))
+	}
 	return &vmspec, errors.New("no suitable infra flavor found for requested node resources, " + strings.Join(reasons, ", "))
 }
 
+// GPU requirements must match exactly to avoid wasting gpu resources
```
levshvarts, line 425: So in case I don't care what type of nvidia gpu I get as long as it has >8G of vram, there is still a hard requirement of the nvidia gpu model. Long term I see it as a good idea to loosen this requirement for example in case of autoprov deploying apps to different sites. What do you think?

levshvarts (commented):
That's a pretty good start for abstracting how we deal with the gpus! Thanks for adding this.
I didn't see any change in the `NvidiaGPUOperatorApp`. Is the assumption for now that the clusters will come with the operator pre-installed?

gainsley:
> I didn't see any change in the `NvidiaGPUOperatorApp`. Is the assumption for now that the clusters will come with the operator pre-installed?

For now yes, I plan to rework the GPU operator code in a later PR.

api/edgeproto/resources.go
```diff
@@ -80,7 +80,7 @@ func (s *KubernetesResources) Validate() error {
 		if err := s.GpuPool.Validate(); err != nil {
 			return fmt.Errorf("gpu pool %s", err)
 		}
-		if s.GpuPool.TotalOptRes == nil || s.GpuPool.TotalOptRes["gpu"] == "" {
+		if len(s.GpuPool.TotalGpus) == 0 && (s.GpuPool.TotalOptRes == nil || s.GpuPool.TotalOptRes["gpu"] == "") {
```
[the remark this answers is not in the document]
gainsley: Yup, there is some intent in the test code in places to support other tags like "nas", but then at the API entry point it's limited only to "gpu" types. It may still be useful in some form to support other tags, but I think it'd be better to rework it as it's really just too confusing to use.

pkg/cloudcommon/deployment.go
```diff
@@ -150,7 +150,7 @@ func isValidKubernetesManifestForResources(objs []runtime.Object, kr *edgeproto.
 		}
 		for j, _ := range template.Spec.Containers {
 			resources := &template.Spec.Containers[j].Resources
-			gpuResName := v1.ResourceName(GPUResourceLimitName)
+			gpuResName := v1.ResourceName(KubernetesNvidiaGPUResource)
```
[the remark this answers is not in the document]
gainsley: Let me fix that to count AMD gpus as well.

pkg/cloudcommon/infraresources.go
```diff
@@ -263,3 +246,52 @@ func ParseOptResVal(resStr string) (string, string, int, error) {
...
+		opt(opts)
+	}
+	for _, gpu := range gpus {
+		if gpu.ModelId == "" {
+			return errors.New("gpu model id cannot be empty")
+		}
+		if gpu.Count == 0 {
+			return errors.New("gpu count cannot be 0")
+		}
+		if gpu.Vendor == "" {
```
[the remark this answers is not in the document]
gainsley: Yeah they're redundant with cloudcommon.ValidateGPUs(), I'll remove them. Unfortunately we can't move cloudcommon.ValidateGPUs() here because it relies on resource definitions from cloudcommon, and cloudcommon already depends on edgeproto.

pkg/controller/cloudlet_resource_check.go
```diff
@@ -35,6 +35,8 @@ type CloudletResCalc struct {
 	cloudletKey *edgeproto.CloudletKey
 	deps        CloudletResCalcDeps
 	usedVals    resspec.ResValMap // cached calculation of used resources in cloudlet
+	skipLB      bool
+	skipAddtl   bool
```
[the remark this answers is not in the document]
gainsley: Makes sense, will do.

pkg/controller/zone_api.go
gainsley: Yup, max is 0 is unlimited. It's the case where the infra is not reporting what the upper limit is.

pkg/k8smgmt/cluster.go
gainsley: Hm, yeah, but then we'd be walking the item.Status.Capacity map twice, since we'd have to then walk it in both GetNodeInfos() and getNodeGPUInfo(). If we wanted to be picky I could just flatten out getNodeGPUInfo() into GetNodeInfos(), as gpu info is potentially spread over both the item.status and item.labels. But I think it's ok for now.

pkg/k8smgmt/cluster.go
gainsley: Sorry, let me fix that comment, what I meant there was multiple GPU types, i.e. a A100 and a T4 on the same node. I believe the operator doesn't support that either currently.

pkg/k8smgmt/cluster.go
gainsley: You're right, I believe I should be using gpu.count instead. I'll fix that.

pkg/platform/azure/azure.go
gainsley: Yup, if they add another set of GPU flavors we'll need to update it.

pkg/resspec/vmspec_matcher.go
gainsley: I agree, long term this could change into something more flexible. Perhaps specifying a GPU family like "ampere" plus a VRAM amount (assuming GPUs can be scheduled with VRAM limits). Right now though I think we have to stick with the specific model id.

gainsley:
@levshvarts thanks for reviewing! PR updated.

levshvarts (approved):
Thanks Jon - looks good!

PR #438, opened 2025-04-08
Fix pod labels issue in kube-state-metrics

vinay03:
### Issues Fixed

`kube-state-metrics` pod, which is responsible for exporting `kube_pod_labels` metric is not including pod labels as attributes while exporting metrics.

### Root Cause: 

kube-state-metrics v2.x has a breaking change which stopped these label attributes from getting attached to `kube_pod_labels` metric.

### Fix
`--metric-labels-allowlist` is an argument to `kube-state-metrics` container which takes in configuration for which labels to be allowed for including into `kube_pod_labels` metric as attributes.

### DIFF
--- a/pkg/cluster-svc/cluster-svc.go
+++ b/pkg/cluster-svc/cluster-svc.go
@@ -91,6 +91,9 @@ defaultRules:
     prometheus: true
     prometheusOperator: true
     time: true
+kube-state-metrics:
+  extraArgs:
+    - --metric-labels-allowlist=pods=[*]
 grafana:
   enabled: false
 alertmanager:


mupasak (approved):

vinay03:
@levshvarts 

Let me know if you need more context on this.

levshvarts (approved):
Looks good. Thanks for fixing this. In general we have a helm chart version set, so I am a bit surprised that kube-state-metrics version got bumped up, but that probably means that for new clusters we don't see metrics.

vinay03:
@levshvarts 
Yes. For new clusters we are not seeing metrics. And this seems to be one of the root causes. Though the metrics issue is not yet solved.

gainsley:
@levshvarts can this be merged?

levshvarts:
> @levshvarts can this be merged?

yes, sorry forgot to merge.

PR #439, opened 2025-04-15
cloudlet managed clusters

gainsley:
This introduces a feature for Cloudlet Managed Clusters, which are clusters that are managed (created/deleted/maintained) by the Cloudlet operator. These clusters can be exposed to our platform via the underlying infra APIs, and registered as ClusterInsts. Once they are registered, our platform can then deploy Applications to them as usual.

To show/register/deregister cloudlet-managed-clusters, three new APIs have been added.

There is no way to "auto-register" existing clusters, as the operator needs to decide how the clusters are exposed - as multi-tenant, reservable, or which organization to allocate them to if they are dedicated to a single tenant.

Currently cloudlet-managed-clusters are only supported for OSM (OpenSource Mano).


pkg/controller/clusterinst_api.go
```diff
@@ -788,121 +789,124 @@ func (s *ClusterInstApi) createClusterInstInternal(cctx *CallContext, in *edgepr
...
-			settings := s.all.settingsApi.Get()
-			if settings.MasterNodeFlavor != "" {
-				masterFlavor := edgeproto.Flavor{}
-				masterFlavorKey := edgeproto.FlavorKey{}
-				masterFlavorKey.Name = settings.MasterNodeFlavor
-				if s.all.flavorApi.store.STMGet(stm, &masterFlavorKey, &masterFlavor) {
-					masterNodeResources = masterFlavor.ToNodeResources()
+			// Handle control (master) node specification
+			// Note for platforms that manage the control nodes themselves
+			// like Azure AKS, etc, we don't need to configure any master
```
levshvarts, line 833: Oh wow - git got sooo confused here :)

gainsley: Yeah it's not very good at figuring out a whole bunch of code just got indented.

pkg/platform/common/infracommon/common-cluster.go
```diff
@@ -0,0 +1,65 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	cmcClustersOut := []*edgeproto.CloudletManagedCluster{}
+	for _, cmcCluster := range cmcClusters {
+		cinfo, found := ecClusters[cmcCluster.Key.Name]
+		if found {
+			if !cinfo.isManagedCluster {
```
levshvarts, line 53: Why not skip edge-cloud managed clusters in the cache traversal?

pkg/platform/common/managedk8s/mk8s-cluster.go
```diff
@@ -200,3 +211,39 @@ func (m *ManagedK8sPlatform) ensureKubeconfig(ctx context.Context, clusterInst *
...
+	if err != nil {
+		return nil, err
+	}
+	getInfraClusterName := func(ci *edgeproto.ClusterInst) string {
+		return m.Provider.NameSanitize(k8smgmt.GetCloudletClusterName(ci))
+	}
+	return infracommon.FilterCloudletManagedClusters(m.CommonPf.PlatformConfig.CloudletKey, cmcClusters, m.caches.ClusterInstCache, getInfraClusterName)
+}
+
+func (m *ManagedK8sPlatform) GetCloudletManagedClusterInfo(ctx context.Context, in *edgeproto.ClusterInst) (*edgeproto.CloudletManagedClusterInfo, error) {
```
levshvarts, line 226: Do we need to spell out the permissions on the managed cluster? It looks like we should at least be able to get nodes, create namespaces, etc.

levshvarts (approved):
Looks good. Thanks for adding this.

pkg/platform/common/infracommon/common-cluster.go
gainsley: That would cause the lookup to fail, which then becomes ambiguous - is it a managed cluster that I already know about, or is it a managed cluster being reported by the infra, but I have no corresponding cluster in the cache for? Let me add some comments to make it more clear what the intent of this function is.

pkg/platform/common/managedk8s/mk8s-cluster.go
gainsley: I think the short answer is we need admin access to the cluster.

PR #441, opened 2025-04-16
rename node to svcnode

gainsley:
This just renames Node to SvcNode. SvcNode represents an instance of one of our services, like the Controller or CCRM. For bare metal, we would like to define a Node as a VM/BareMetal machine. So we are renaming the existing Node to SvcNode. There are no logic changes here, just renames.


levshvarts (approved):
Is `nodemgr` now ambiguous as well?

gainsley:
> Is `nodemgr` now ambiguous as well?

Good call I should rename that too.

PR #443, opened 2025-05-01
Crm debug to get logs

levshvarts:
### Description
In the restricted cloudlet environment it's difficult to get the logs from crm. This change adds a new command
```sh
mcctl debug rundebug cmd=crmlogs args="<grep args>" cloudlet=<cloudletname>
```
It runs `docker logs crmserver` on the platform vm. Command also takes a string as an arg, which will be used to grep the logfile for(`docker logs crmserver 2>&1 | grep "<args>"`).
NOTE: cloudlet name is mandatory, due to the potential of the command to return a very large amounts of data.

In addition I added more logs for when the k8s cluster create fails.


pkg/gencmd/notice.cmd.go
```diff
@@ -34,7 +34,7 @@ var NoticeAliasArgs = []string{}
...
...ed protocol buffer message. The last segment of the URLs path must represent the fully qualified name of the type (as in `path/google.protobuf.Duration`). The name should be in a canonical form (e.g., leading . is not accepted). In practice, teams usually precompile into the binary all types that they expect it to use in the context of Any. However, for URLs which use the scheme `http`, `https`, or no scheme, one can optionally set up a type server that maps type URLs to message definitions as follows: * If no scheme is provided, `https` is assumed. * An HTTP GET on the URL must yield a [google.protobuf.Type][]   value in binary format, or produce an error. * Applications are allowed to cache lookup results based on the   URL, or have them precompiled into a binary to avoid any   lookup. Therefore, binary compatibility needs to be preserved   on changes to types. (Use versioned type names to manage   breaking changes.) Note: this functionality is not currently available in the official protobuf release, and it is not used for type URLs beginning with type.googleapis.com. Schemes other than `http`, `https` (or the empty scheme) might be used with implementation specific semantics.",
```
gainsley, line 37: I think this change is because you and I have different versions of something - not sure exactly what.

gainsley (approved):

pkg/gencmd/notice.cmd.go
levshvarts: Ah, yeah, I'll clean it up.

